PR #55, opened 2014-08-29
Fix unauthenticated mode.

kryzthov:
When no htaccess files is specified, uploads always fail with a 403.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -27,7 +27,9 @@ config = configuration()
 
 
 def validate_user(username, password):
-    if config.htpasswdfile is not None:
+    if config.htpasswdfile is None:
+        return True
+    else:
         config.htpasswdfile.load_if_changed()
         return config.htpasswdfile.check_password(username, password)
 


ankostis:
@schmir currently to enable package-uploads requires the `htpasswd` command, and that makes it hard for Windows-users to use this feature. 

Would it make sense to revert this decision, and support uploads by default?
With some "explicit" syntax like ``-P .`` it maybe obvious to the users that they are overriding security.

schmir:
I think it should be rather easy to implement something like htpasswd with passlib. But are there even any windows users that need that feature?

ankostis:
Currently i do :-(
And have to copy around linux-generated files.

If you agree we can schedule this for the next release. 
(by the way i'm releasing 1.1.7 in a couple of minutes ;-)

schmir:
ankostis, you really don't need my agreement on anything. If you feel like something should be changed, just do it if your co-maintainers are ok with it. It's your show now. You're doing the work, you're going to decide things!

And thanks for taking over!

ankostis:
I don't want to consume more of your time.
I do appreciate your way of thinking, as this has been imprinted into this project.  
But since it is not yet possible to completely transfer brain-snapshots :-) i would still like to hear some background info on your previous decisions.
In any case, to avoid further burden on you, i will reserve my questions for really blocking issues.

Thanks again for your cooperation.

awestendorf:
This could also be solved by changing the default in [main](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/core.py#L296) to:

```python
def main(argv=None):
    ...
    authenticated = []
```

It would be really great to fix this, because it's a major pain on any system. Without TLS support, forcing authentication in pypiserver is of little value as no user will want their passwords sent in plain text. I plan to run nginx in front of it, and I assume many others will want this flexibility as well. I was going to write a PR and then found this one.

The change in this PR, the one I suggested, or both, would solve the problem.

PR #61, opened 2014-11-14
Possible to redefine server's welcome-page with cmd-line option (2).

ankostis:
* Extract http welcome-msg(/) in a separate ASCII file.
* Add cmd-line option --welcome for changing the welcome-file and update docs.
* Add 2 TCs, for cmd-line option and / http-req.


petri:
The welcome msg should NOT be ascii but encoded as UTF-8 to support international users, please...

ankostis:
I made it consume utf-8 and provide an in-code message if reading failed (for standalone).
Also squashed and rebased changes on latest master.

PR #70, opened 2015-02-15
Updated PR #59, test on py34 also

ankostis:


ankostis:
Merged #56 with current master, and added PY34 on tox.

PR #72, opened 2015-02-16
Fix some PEP8 conventions

ankostis:


ankostis:
Transfered in @ankostis fork.

PR #78, opened 2015-02-20
FIX #77 XSS on Welcome page

### LINKED ISSUE #77: Cross Site Scripting Vulnerability
Versions up to and including 1.1.7 contain a cross site scripting vulnerability.  I acknowledge this is not likely to ever be a problem since the normal usage of the server is from pip rather than a web browser.  I only note it because my server was flagged as vulnerable by a Nessus scan.

The vulnerability can be demonstrated by entering the following in a web browser that does **not** filter XSS (on modern browsers, you'll probably have to disable automatic XSS filtering to see the exploit)
`http://my.pypi-server.com:8080/?<script>alert("Vulnerable!");</script>`

If successful, this should pop up an alert window,

ankostis:


ankostis:
@schmir @ror6ax Would you mid reviewing & merging this PR?

I'm certain that i would need some more attention to the test-cases.
Note that i've also update the welcome-message TCs with this PR.

PR #85, opened 2015-07-17
More deterministic sort of packages

asottile:
We run an internal pypi server and have both sdists and wheels on it.  When refreshing they reorder because the sort only sorts on version number.  This change makes that sort more deterministic.


dnephin:
+1

PR #95, opened 2015-10-27
Add missing template loop terminators

bmflynn:
Fixes redundant `</body></html>` after each item in lists.


ankostis:
Thank you @bmflynn, rebased as 1d9e8dfd.

PR #98, opened 2015-12-11
Fixes #91

### LINKED ISSUE #91: support for setup.py register command is broken
I have fresh pypiserver 1.1.7 running on current Debian Stable, on Python2.7.

For some reason, running "python setup.py sdist register upload" works, but just "setup.py register" does not (locally):

    Server response (500): <urlopen error [Errno 104] Connection reset by peer>

Or from a remote OSX system:

    Server response (500): <urlopen error [Errno 32] Broken pipe>

... and likewise, using jarn.mkrelease (tool to automate releases) from the OSX system to do uploads fails:
(it does register - sdist - upload in three separate HTTP calls, I guess)

    running register
    Registering z3c.form to http://pypi.vidamin.com:8765
    Server response (500): <urlopen error [Errno 32] Broken pipe>
    ERROR: register failed

In all these cases, the pypiserver log shows just:

    "POST / HTTP/1.1" 401 713 "-" "Python-urllib/2.7"

Which is not very helpful...

pypiserver should accept the register call on its own, even if just to satisfy clients (and users) that try to do "setup.py register" on its own without other commands. My suggestion would be to simply silently accept the call, and do nothing (except log the event perhaps).

Note: I suspect this issue may be behind a few other issue reports about "random" 401 or 500 errors, as well... I thought for a long time I was seeing the same issue.

At the very least, it should be documented that doing just "setup.py register" does not work.

Tythos:
Keyword argument 'header' will not be written to the corresponding header field in the 401 response. This prevents iterative HTTP requests (i.e., urllib2.HTTPBasicAuthHandler as used for distutils.command.register, as opposed to directly specifying the 'Authorization' header value in distutils.command.upload) from appropriately responding to a misplaced "www-authenticate" response header. To appropriately transcribe 'www-authenticate' when it passes from bottle.HTTPError.__init__ as an **option field to bottle.HTTPResponse__init__ as mapped to the 'headers' field, this argument was re-named.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -55,7 +55,7 @@ class auth(object):
             if self.action in config.authenticated:
                 if not request.auth or request.auth[1] is None:
                     raise HTTPError(
-                        401, header={"WWW-Authenticate": 'Basic realm="pypi"'})
+                        401, headers={"WWW-Authenticate": 'Basic realm="pypi"'})
                 if not validate_user(*request.auth):
                     raise HTTPError(403)
             return method(*args, **kwargs)


ankostis:
Excellent.
As you may have noticed i'm a bit strained on the availability for this project.
I will try to release a fixed final release within a reasonable timeframe.

Thank you.

petri:
+1 for merging and new release (I've used a snapshot of this code in production for a while now without problems)

ankostis:
Merged reworded as 98c654d0b.

Issue #25, opened 2012-11-23
support voor virtual hosts


woutervh:
I have pypiserver running on apache/mod_wsgi, served under "/pypi"

but the some links on the homepage are incorrect 
because they don not take into account request.fullpath

in 
   58 @app.route('/')
   59 def root():

 "packages" point to "/packages" instead of "/pypi/packages"
 "simple" point to "/packages" instead of "/pypi/simple"

PR #100, opened 2016-01-02
Fix PACKAGES and SIMPLE addresses

eugulixes:
I use Nginx as a reverse proxy. For example, if the welcome page of my repository is available at https://myserver/pypi, then the `here` and `simple` links on the page will be pointing to `https://myserver/packages/` and `https://myserver/simple/` respectively. It's wrong.  I propose a small patch which fixes it.

### DIFF
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -36,6 +36,7 @@ def default_config(
         host = "0.0.0.0",
         port = 8080,
         server = DEFAULT_SERVER,
+        url_prefix=None,
         redirect_to_fallback = True,
         fallback_url = None,
         authenticated = ['update'],
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -78,6 +78,10 @@ def usage():
       default is to use "auto" which chooses one of paste, cherrypy,
       twisted or wsgiref.
 
+    -f, --url-prefix PREFIX
+      in case of using nginx with custom path you have to use PREFIX to
+      override links
+
     -r, --root PACKAGES_DIRECTORY
       [deprecated] serve packages from PACKAGES_DIRECTORY
 
@@ -162,13 +166,14 @@ def main(argv=None):
     update_stable_only = True,
 
     try:
-        opts, roots = getopt.getopt(argv[1:], "i:p:a:r:d:P:Uuvxoh", [
+        opts, roots = getopt.getopt(argv[1:], "i:p:a:r:d:P:f:Uuvxoh", [
             "interface=",
             "passwords=",
             "authenticate=",
             "port=",
             "root=",
             "server=",
+            "url-prefix=",
             "fallback-url=",
             "disable-fallback",
             "overwrite",
@@ -216,6 +221,8 @@ def main(argv=None):
             c.fallback_url = v
         elif k == "--server":
             c.server = v
+        elif k in ("-f", "--url-prefix"):
+            c.url_prefix = v
         elif k == "--welcome":
             c.welcome_file = v
         elif k == "--version":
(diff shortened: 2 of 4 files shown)


ankostis:
Many TCs fail :-(

ankostis:
But what it is strange is that I don't have this problem with apache as a front-end.
@eugulixes Could it be a Ngix config issue?

eugulixes:
It was quite stupid of me not to run the tests before sending the pull request.
My problem is very similar to [this one](https://github.com/pypiserver/pypiserver/issues/25) but still a bit different. When I proxy `https://myserver/npm` to `http://127.0.0.1:8080`, request.[fullpath](http://bottlepy.org/docs/dev/api.html#bottle.BaseRequest.fullpath) is empty, logically I guess. For example, [Sinopia](https://github.com/rlidwka/sinopia), the npm repository server , has the `url_prefix` option in its config file. The comment to the option is the following: «if you use nginx with custom path, use this to override links». I think we have to provide something like this. Check out my latest commit please.

ankostis:
I would prefer not to add an ad-hoc solution just for the landing-page links, that might break something else further down the road; a reverse proxied *pypiserver* SHOULD work flawlessly without needing to manually set the request-url.

I think the problem is that Ngnix does not support the apache's(?) `X-forward-host` header.
There are many suggestions how to workaorund that: 
  https://www.google.com/search?q=nginx+reverse-proxy+"X-Forwarded-Host".

I would appreciate if you reported here what you did, so that we can add a note in the documents.

eugulixes:
I understand your conservatism. However everything is supposed to remain the same if you don't specify the `--url-proxy` option. As you have mentioned, you use Apache as a front-end, so you could give feedback as a user. For example, if you don't notice any regression after update then everything works as intended.
I use PyPIServer along with Sinopia in which the problem has already been solved (described in my previous message). Since I like the solution, I added it to PyPIServer. I run the latest stable release of PyPIServer in a docker container, managing a set of patches via [quilt](https://wiki.debian.org/UsingQuilt), so it's not a problem for me to rebuild the container with or without this patch as many times as I want. Unfortunately this information is not useful for the PyPIServer documentation.

ankostis:
Indeed, you are right in how you see this, from a point of view.
Everything is supposed to remain the same if you *don't use it*.

Yet, if people choose to workaround the problem by *using this flag*, 
instead of properly configuring their proxy, and if that causes other sides-effects 
(because, in the end, they have a *miss-configured proxy),
it would translate to more invalid bug-requests, help issues, more workaround-fixing and possibly new enhancements in the wrong direction.

So in the very end, not everything would be exactly the same; some people may have spared their time.  (I hope it is not too egoistic all this).

eugulixes:
Fair enough.

ankostis:
Thank you for your cooperation.

kujyp:
> I would prefer not to add an ad-hoc solution just for the landing-page links, that might break something else further down the road; a reverse proxied _pypiserver_ SHOULD work flawlessly without needing to manually set the request-url.
> 
> I think the problem is that Ngnix does not support the apache's(?) `X-forward-host` header.
> There are many suggestions how to workaorund that:
> [https://www.google.com/search?q=nginx+reverse-proxy+"X-Forwarded-Host](https://www.google.com/search?q=nginx+reverse-proxy+%22X-Forwarded-Host)".
> 
> I would appreciate if you reported here what you did, so that we can add a note in the documents.

@ankostis 
There's a typo.
I think the problem is that Ngnix does not support the apache's(?) `X-forward-host` header.
-> `X-forwarded-host`

mplanchard:
Are you having the same issue as the OP? 

Are you using a reverse proxy with nginx as described [here](https://github.com/pypiserver/pypiserver#behind-a-reverse-proxy)?

Please go ahead and open a new issue with some more details, so we can try to get any issues you're having resolved. A fair bit has changed since 2016 :)

Issue #53, opened 2014-06-03
Provide hash in the link for local packages


Ivoz:
pip 1.5 and greater would like secure links by default, which by implementation means the link provides a hash of the package in the url. This provides some form of verification of correct file download at least.

You can see some of the [api docs at warehouse](http://warehouse.readthedocs.org/en/latest/api-reference/legacy/#get--simple--project--) how this is done.

It would be great for pypiserver's links to provide this for pip; I believe this would mean you no longer have to pass `--allow-insecure`.

ankostis:
I may have gotten this wrong, but why is a DB needed for pip's secure-links?
Isn't it enough for pypiserver to check that the requested hash matches the hosted or fail-through package's hash?

ror6ax:
Well, remaking hash every time it's requested is a bit overhead, right?
I'll finish this in nearest future.

ankostis:
My feeling is that current CPUs have more than enough juice for such numerical tasks.
Of course this may not apply for embedded ones.

ankostis:
Note that the hashes contained on the link-fragments served by PyPI are NOT used by `pip`: https://pip.pypa.io/en/latest/reference/pip_install/#hashes-from-pypi

Ivoz:
:thumb: nice work ankostis

PR #101, opened 2016-01-04
Serve tens of thousands of packages efficiently

virtuald:
pypiserver has worked great for us over the last 2 years, but over time we've accumulated 25000 individual package files (mostly artifacts from automatic builds) and pypiserver has really slowed down for us.

This PR adds some significant optimizations for this particular use case -- in particular, it holds the list of packages in memory so we don't have to scan all 25000 packages each time pip queries the server (which is a few times per package). It uses the `watchdog` package to invalidate the cache, and will not activate the cache if it is not installed.

Another significant speedup we found was using an nginx proxy, and adding the following to make nginx serve the packages directly by pointing the `/packages/` location to its parent directory:
```
  location /packages/ {
    root /path/to/packages/parentdir;
  }
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -368,6 +368,24 @@ Specifying an additional `-u` option will also allow alpha, beta and
 release candidates to be downloaded. Without this option these
 releases won't be considered.
 
+Serving thousands of packages
+-----------------------------
+
+By default, pypiserver scans the entire packages directory each time an
+incoming HTTP request occurs. This isn't a problem for a small number of
+packages, but causes noticeable slowdowns when serving thousands or tens
+of thousands of packages.
+
+If you run into this problem, significant speedups can be gained by enabling
+pypiserver's directory caching functionality. The only requirement is to
+install the ``watchdog`` package, or it can be installed by installing
+``pypiserver`` using the ``cache`` extras option::
+
+  pip install pypi-server[cache]
+
+If you are using a static webserver such as Apache or Nginx as a proxy for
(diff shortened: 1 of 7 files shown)


virtuald:
Fixed the tests; the one annoying thing at the moment is that because of the fix applied as a result of #53, generating the hashes for 1000+ packages is still pretty time consuming. I did cache those also, but to really get this where it needs to be the listdir cache needs to be a bit more sophisticated and only invalidate things that changed, instead of the whole cache.

ankostis:
@virtuald grateful for your quality PR.

Noting that the latest release [`v1.1.9`](/pypiserver/pypiserver/releases/tag/v1.1.9) will inflict even more performance hit due to repetitive hashing from #53.
I would appreciate any feedback on the actual impact of hashinh in your case

virtuald:
As noted above, the hashing is pretty annoying and takes a fair amount of extra time. But, since I'm caching the hashes too, this PR is still a huge improvement over where we were.

ankostis:
@virtuald are you interested to invite you into *pypiserver* devs?

virtuald:
Thanks for the offer, but I'll have to decline as I already have too many side projects that I don't put enough time into. :)

dandesousa:
I was about to suggest this exact feature and implement a PR. Glad I checked here first. +1 on getting this merged and made available in Pypi.

virtuald:
We've been using it for two weeks without any issues that I'm aware of.

PR #111, opened 2016-01-27
Audit Log

iiSeymour:
Simply adds the username of the user when a package is uploaded.


mplanchard:
Python 3.3 tests failed initially, but succeeded when I re-ran. Note that the `'{}'.format(variable)` syntax does not work in Python 2.6, which requires numbers in the brackets. However, we are also dropping support for 2.6. Since the log functions won't run unless logging is enabled, performance impact should be negligible.

ankostis:
Accepted as fdf5accd755.

Thanks!

PR #113, opened 2016-02-16
Bug fix for overwriting existing files without the overwrite flag bei…

blade2005:
Bug fix for overwriting existing files without the overwrite flag being set to true.

Looks like a simple oversite where a raise should have been done.

The current log output looks like 

<LocalRequest: POST http://10.1.10.155:8092/>
Cannot upload 'testpackage-2.0.2.tar.gz' since it already exists! 
  You may start server with `--overwrite` option. 
Stored 'testpackage-2.0.2.tar.gz'.
200 OK


blade2005:
Hrm, I'm not seeing any reason why this would cause 400 error's on 3.3

PR #116, opened 2016-03-04
Add a separate cache for file digests

virtuald:
Turns out, those file digests get really annoying if you have a hundred or so packages that are 40mb each (don't ask... ). This improves the cache and makes that less annoying.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -407,6 +407,9 @@ packages-directly directly (take care not to expose "sensitive" files)::
       root /path/to/packages/parentdir;
     }
 
+If you have packages that are very large, you may find it helpful to
+disable hashing of files (set `--hash-algo=off`, or `hash_algo=None` when
+using wsgi).
 
 Using a different WSGI server
 -----------------------------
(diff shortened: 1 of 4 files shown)


ankostis:
Nice,
Just describe a bit more in the commit message and/or here why #101 was not enough, and what does this differently.
I don't have any experience with such heavy loads.

virtuald:
Updated the commit, should work now...

virtuald:
Also added a fix for --hash-algo=off in here, in case someone decides that approach is better for them.

ankostis:
Very well!

(are you still insisting not to participate to `pypiserver` team?)

virtuald:
Yep. Just because I don't want to be on the team doesn't mean that I won't fix issues that affect me. :)

mplanchard:
Can we get unit tests for the new caching functionality?

If not, I'll look into adding some this weekend.

ankostis:
I was also not accustomed to pytest when I took over this project - and I think the setup of the fixtures is not optimal, but i still gave it a try and it is not completely worthless framework.  
Anyway, you mau add unittests and I ll try to convert them to pytests.

mplanchard:
@ankostis Sorry, when I say unit tests, I meant pytests. I use "unit test" as a generic catchall term for unit tests, pytests, nose tests, etc. Pytest is fantastic, and any tests we add should use the pytest framework.

PR #126, opened 2016-03-21
Fix typo, the package name is pypiserver

harcher81:
Fix typo, the package name in pypi is pypiserver and not pypi-server.


mplanchard:
Good catch! Thanks very much.

PR #132, opened 2016-04-30
Add matrix auth

khornberg:
This allows multiple users to have differing levels of permissions. In particular, a read-only (list, download) user and a write (upload) user.

I've provided no changes to the documentation as I expect the specific syntax may need to change.

### DIFF
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -15,6 +15,7 @@ import textwrap
 import warnings
 
 import functools as ft
+import ast
 
 
 warnings.filterwarnings("ignore", "Python 2.5 support may be dropped in future versions of Bottle")
@@ -201,8 +202,16 @@ def main(argv=None):
                 sys.exit("Invalid port(%r) due to: %s" % (v, err))
         elif k in ("-a", "--authenticate"):
             if '{' in v:
-                import ast
-                v = ast.literal_eval(v)
+                try:
+                    v = ast.literal_eval(v)
+                except SyntaxError:
+                    message = 'Could not parse auth string %s! Please ensure string is correctly formatted.' % v
+                    print(message)
+                    sys.exit(message)
+                if (not isinstance(v, dict) or not all([isinstance(i, list) for i in v.values()])):
+                    message = 'Matrix auth string must be a dict of lists. Please see the README for details.'
+                    print(message)
+                    sys.exit(message)
             if isinstance(v, dict):
                 c.authenticated = {}
                 for user in v:
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -197,17 +197,33 @@ def main(argv=None):
                 err = sys.exc_info()[1]
                 sys.exit("Invalid port(%r) due to: %s" % (v, err))
         elif k in ("-a", "--authenticate"):
-            c.authenticated = [a.lower()
-                               for a in re.split("[, ]+", v.strip(" ,"))
-                               if a]
-            if c.authenticated == ['.']:
-                c.authenticated = []
+            if '{' in v:
+                import ast
+                v = ast.literal_eval(v)
+            if isinstance(v, dict):
+                c.authenticated = {}
+                for user in v:
+                    c.authenticated[user] = [a.lower() for a in v[user] if a]
+                    if c.authenticated[user] == ['.']:
+                        c.authenticated[user] = []
+                    else:
+                        actions = ("list", "download", "update")
+                        for a in c.authenticated[user]:
+                            if a not in actions:
+                                errmsg = "Action '%s' for option `%s` not one of %s!"
+                                sys.exit(errmsg % (a, k, actions))
             else:
-                actions = ("list", "download", "update")
-                for a in c.authenticated:
-                    if a not in actions:
-                        errmsg = "Action '%s' for option `%s` not one of %s!"
-                        sys.exit(errmsg % (a, k, actions))
+                c.authenticated = [a.lower()
+                                   for a in re.split("[, ]+", v.strip(" ,"))
+                                   if a]
+                if c.authenticated == ['.']:
+                    c.authenticated = []
+                else:
+                    actions = ("list", "download", "update")
+                    for a in c.authenticated:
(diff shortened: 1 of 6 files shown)


bibby:
Looking forward to this.
One quick thought, since you mentioned syntax. The config offers an alternative `auther`, should one prefer an authentication method other than `htpasswd`. The `htdigest` format seems like a good fit to match a user with their capabilities.

khornberg:
@bibby I'm not sure. I've tried to keep this change as minimum as possible.

khornberg:
As for the failing ci status, I've not worked with Travis CI before so I don't know how to fix it. The tests pass on py 3.4.4 locally running `py.test`. Looking at the output it seems that the environments are different from my machine because one of the server tests is expecting a password to be inputed.

bibby:
@khornberg , changing little one reason I suggest it, tbh.
I could use an ldap auther for my org, and using this `--auther` hook looks like the place to do it.

The [default auther](https://github.com/pypiserver/pypiserver/blob/b1b36f3a133d6027e2db86de78ddcdc95c94e2e6/pypiserver/core.py#L47-L53) could even be rolled into a separate module as one of multiple auth strategies.

In my case, LDAP brings with it a lot of depencies and config pain; I may lookinto an making an outside package, `pypiserver-auth-ldap` or the like.

dpkp:
See #141 re test failures

mplanchard:
@khornberg could you rebase this branch against the current upstream master to resolve conflicts?

khornberg:
@mplanchard done

mplanchard:
@khornberg Thanks! Would you mind adding some documentation for usage in the README, perhaps as a new section under `detailed usage`?

pypiserver/__main__.py
```diff
@@ -195,17 +195,33 @@ def main(argv=None):
...
                 sys.exit("Invalid port(%r) due to: %s" % (v, err))
         elif k in ("-a", "--authenticate"):
-            c.authenticated = [a.lower()
-                               for a in re.split("[, ]+", v.strip(" ,"))
-                               if a]
-            if c.authenticated == ['.']:
-                c.authenticated = []
+            if '{' in v:
+                import ast
+                v = ast.literal_eval(v)
```
mplanchard: Given this comes from user input, we may want to try/except `ValueError` and give a useful error message if their input doesn't parse. It's probably also worth doing something like the following to ensure that, if it parsed correctly, that it's the correct format:

```python
try:
  v = ast.literal_eval(v)
except ValueError:
  sys.exit('Could not parse auth string %s! Please ensure string is '
           'correctly formatted.' % v)
if (not isinstance(v, dict)
        or not all([isinstance(i, list) for i in v.values()])):
    sys.exit('Matrix auth string must be a dict of lists. Please see the '
             'README for details.')
````
We could then also roll the next block of logic `if isinstance(v, dict):` into the else for that statement, keeping it all within the if block starting with `if '{' in v:`, since there's no other case in which v will be a dict anyway.

**Comment by mplanchard** [comment]
Also @bibby you may want to check out #149 - the author mentioned originally developing the feature for ldap.

**Comment by blade2005** [comment]
With ldap/pam authentication it validates if the user/pass is correct. #149 wasn't done from a restriction on group but validating users. To validate groups I think #132  is on the right track. A dict of lists, where under htpasswd auth the it's treated as users. Under pam auth, users or groups. 

Because auther could be set as a custom callable in the api we would also need to add a second callable to evaluate if the user/group has access based on the authenticate option, for htpasswd this is quite simple but for the pam method this also requires getting the user groups and checking that as well. 

In the interest of simplicity I'm not 100% sure this type of change suits the idea of a simple pypi server. @mplanchard or @ankostis what are your thoughts on this?

**Comment by blade2005** [comment]
```
def check_user_group(authenticated, username, request):
    """authenticated should be the full list from opts. request should be update,list,download

    http://stackoverflow.com/questions/9323834/python-how-to-get-group-ids-of-one-username-like-id-gn"""
    import grp, pwd 
    groups = [g.gr_name for g in grp.getgrall() if username in g.gr_mem]
    gid = pwd.getpwnam(username).pw_gid
    groups.append(grp.getgrgid(gid).gr_name)
    if username in authenticated[request] or [group for group in groups if group in authenticated[request]]:
        return True
    else:
        return False
```

Completed untested code but should work.

khornberg:
Yes, I'll try to get the docs next week.

ankostis:
Let's wait another release-cycle for this.  Maybe it remains simple after all.

pypiserver/__main__.py
khornberg: @mplanchard I don't follow
> We could then also roll the next block of logic if isinstance(v, dict): into the else for that statement, keeping it all within the if block starting with if '{' in v:, since there's no other case in which v will be a dict anyway.

mplanchard:
This is excellent work, and I don't want to see it languish. @khornberg, would you mind rebasing onto master to resolve the merge conflict?

khornberg:
Not at all

mplanchard:
Looks like we're getting some errors on several tests in the `test_server.py` area, mostly due to differing argument numbers.

mplanchard:
Hm. This might actually be due to something else, as the same issue just popped up in #156; I'll check into it.

mplanchard:
Test failures were due to a newly release version of `twine`. Once #158 is merged, we should be good to go.

wasbazi:
This change would be really helpful for a use case I have, anything I can do to help get these changes merged?

blade2005:
@khornberg can you rebase to include the changes from #158

khornberg:
I think so at some point

Kyle

mplanchard:
I just re-triggered the test build, and it looks like the issues with `twine` are resolved with our version pin. However, we are running into other failures when testing the standalone script. I would like to see if a rebase solves the issue. If you don't mind @khornberg, please @ me when you rebase so I am reminded to check on the build status.

khornberg:
@mplanchard rebased

mplanchard:
Awesome. Merging.

ankostis:
Should become a standlone plugin.

Issue #139, opened 2016-05-13
Packages with dot no longer install(?)


JonasT:
It appears pypiserver is affected by this issue: https://github.com/pypa/pip/issues/3666

At least I can't install my project anymore with the latest pip, and I have dots in all package names.. (pip complains about no such package/no distributions being found)

sbrandtb:
I'm running into this issue, too. I tested the version from master. It's fixed there, but painfully slow

dpkp:
We've deployed the 503 patch internally (PR 140 / master) to general success. Do you have any details on the slowdown you are seeing?

sbrandtb:
We decided to switch to devpi internally because we need pypi-caching, so no chance to check why the version from master was hanging. Probably some DSN lookup or other timeout?

Still, keep up the good work. Pypi-server served us for two years without issues :) Still, I would recommend releasing the latest master because this bug is a critical issue here and probably also in other environments.

ankostis:
Indeed we have to bake a new release the soonest.

JonasT:
This still seems to be broken for me even with latest git master cloned from today (pypiserver version reported as 1.2.0.dev1)

```
pip3 install --extra-index http://172.17.0.1:8765/ daphnetf.charon
Downloading/unpacking daphnetf.charon
http://172.17.0.1:8765/daphnetf.charon/ uses an insecure transport scheme (http). Consider using https if 172.17.0.1:8765 has it available
Could not find a version that satisfies the requirement daphnetf.charon (from versions: 16.4.0.dev0)
Cleaning up...
No distributions matching the version for daphnetf.charon
Storing debug log for failure in /root/.pip/pip.log
```
... with the package in question definitely being in the package dir of pypiserver:
```['daphnetf.charon-16.4.0.dev0.tar.gz', 'daphnetf.hermes-0.0.1.tar.gz', 'daphnetf.gaia-0.0.1.tar.gz']```

However, pip also removed the --allow-external in the same fateful update that broke pypiserver for me, so I don't even know if my pip arguments are still supposed to be working like that. Any help would be *greatly* appreciated, because I find pip documentation on how to install from external sources to be notoriously bad... (which hasn't gotten better by half of the internet examples which used --allow-external now being no longer working as before)

dpkp:
Try adding --trusted-host <your pypiserver IP/hostname> to pip install.
This is required by pip >=8 when not using https

ankostis:
[...]

Issue #38, opened 2013-04-27
PyPI backend and redirections


gracinet:
Hi, thanks for pypiserver, simple and useful !

Here's a tricky problem when running pypiserver with fallback to a mirror index, not sure if it should be fixed within pypiserver or not.
See also https://bitbucket.org/loewis/pep381client/issue/22

The main PyPI server does some 301 redirections (to fix some mispellings). That's not something a mirror can easily reproduce (see above cited issue on pep381client).
The combination of pypiserver and a mirror leads to installation errors. 
Fixing one's spellings is not practically possible, because they occur a lot through dependencies.

### Reproduction

Assume for instance that pypiserver does not hold packages for 'babel'.

Client applications (such as pip) try e.g., '/simple/babel' (proper spelling is 'Babel')

- On the main PyPI, they get a 301 response to '/simple/Babel' -> ok

- On a PyPI mirror, they get a 404, then they fetch the whole '/simple/' page, analyse it,   find 'Babel' in it, correct the spelling and retry -> inefficient, but ok

- On pypiserver with fallback on a mirror, they get:
  - a 303 to the mirror ('/simple/babel')
  - then, a 404 from the mirror
  - then they fetch the '/simple' page *from pypiserver*, which does not list the wished distribution ('Babel') -> ERROR

### Workarounds 

My workaround for now is to setup 301 redirections for commonly mispelled distributions in the mirror as well (in our case, the mirror is ours), but that's a maintenance burden and does not scale.
 
Another possibility is to manually add commonly mispelled distributions in pypiserver and use '-U'. That's insatisfactoy for the same reasons.

### Conclusion

Mispellings in dependencies are really common, since many distribution authors don't even realize their spelling is wrong. Any idea ?

ankostis:
@gracinet Do you think that a workaround would be to invoke `pip` with ``--extra-index-url <pypiserver_url>`` instead of ``--index-url <pypiserver_url>``?

gracinet:
@ankostis, nice to see some new activity around pypiserver !
About your suggestion, I'm not sure and I don't have an easy way to test it anymore : we dropped the pep381 mirror instance completely since then, the main PyPI became much more robust shortly afterwards (that's not the only problem we had with pep381)

If that helps you, I can try and reproduce again, then tell about the suggestion. 

[...]

PR #140, opened 2016-05-15
PEP 503 Package Name Normalization and URL Redirects

dpkp:
PEP 503 addresses #139 (and also #38, which was I believe actual motivation of the PEP...) .

This PR makes the following changes to bring pypiserver up to spec:
- URLs are redirected unless they end in `/` (expect packages themselves).
- Package names are normalized: lowercase with all runs of the characters `.` , `-` , or `_` replaced with a single `-` character.
- The simple index only lists normalized package names.
- Any request for a non-normalized package name is redirected to the normalized name.

Note that pip >=8.1.2 will only request normalized package names. Apparently this change was intended to be in 8.0.0, but an implementation bug did not apply normalization correctly so that packages with dots (`.`) were requested unnormalized.

Tests added / updated.

### DIFF
--- a/pypiserver/core.py
+++ b/pypiserver/core.py
@@ -181,7 +181,8 @@ def guess_pkgname_and_version(path):
 
 
 def normalize_pkgname(name):
-    return name.lower().replace("-", "_")
+    """Perform PEP 503 normalization"""
+    return re.sub(r"[-_.]+", "-", name).lower()
 
 
 def is_allowed_path(path_part):
@@ -258,23 +259,11 @@ def find_packages(pkgs, prefix=""):
 
 
 def get_prefixes(pkgs):
-    pkgnames = set()
     normalized_pkgnames = set()
-    eggs = set()
-
     for x in pkgs:
         if x.pkgname:
-            if x.relfn.endswith(".egg"):
-                eggs.add(x.pkgname)
-            else:
-                pkgnames.add(x.pkgname)
-                normalized_pkgnames.add(x.pkgname_norm)
-
-    for x in eggs:
-        if x not in normalized_pkgnames:
-            pkgnames.add(x)
-
-    return pkgnames
+            normalized_pkgnames.add(x.pkgname_norm)
+    return normalized_pkgnames
 
 
 def exists(root, filename):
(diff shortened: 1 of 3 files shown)


dpkp:
Test failures are unrelated. See #141

dpkp:
thoughts on this?

mplanchard:
Hi @dpkp - super sorry for the slow responses here. My day job's been rather hectic of late. I merged #141 and am re-running the tests for this change. I'm diving into the code now.

pypiserver/core.py
```diff
@@ -181,7 +181,8 @@ def guess_pkgname_and_version(path):
 
 
 def normalize_pkgname(name):
-    return name.lower().replace("-", "_")
+    """Perform PEP 503 normalization"""
+    return re.sub(r"[-_.]+", "-", name).lower()
```
mplanchard: Suggestion: use `pkg_resources.safe_name().lower()` instead, since `pkg_resources` should stay up to date with any future PEP changes

dpkp:
No problemo -- I totally understand. Maintaining an open source project in your spare time is a lot of work. Especially if it gets popular. Thanks for taking a look.

pypiserver/core.py
dpkp: pkg_resources is part of setuptools though, and I think setuptools is best left as a build-time dependency. Indeed, the wheel format is intended to more cleanly separate build-time bits from run-time.

mplanchard: Makes sense.

mplanchard:
Looks good! Thanks especially for including tests.

dpkp:
Thanks! Worth tagging/publishing a new release to pypi?

mplanchard:
Probably! There are a few other outstanding PRs and issues. I'd like to go through them prior to tagging a new release, but I will hopefully have one out by the time the weekend is through!

clarkevans:
Thank you for your lovely pypiserver.  It'd be great if this particular fix could be tagged to make it compatible with recent versions of pip.

ankostis:
@dpkp of course you are welcome to the admin team if you feel like continuing contributing - as you probably have understand by now, we need to increase the response time this project.

PR #141, opened 2016-05-15
upload test requires pypirc to avoid password prompt

dpkp:
Test suite hangs because an upload test prompts for a user-supplied password. This PR uses the pypirc context manager to supply a default password.

See:
https://travis-ci.org/pypiserver/pypiserver/builds/129511699
https://travis-ci.org/pypiserver/pypiserver/builds/126811063
https://travis-ci.org/pypiserver/pypiserver/builds/128599164
https://travis-ci.org/pypiserver/pypiserver/builds/130325284

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -235,14 +235,23 @@ def pypirc_file(txt):
 @pytest.mark.parametrize("pkg_frmt", ['bdist', 'bdist_wheel'])
 def test_setuptoolsUpload_open(empty_packdir, port, project, package,
                                pkg_frmt):
-    with new_server(empty_packdir, port):
-        with chdir(project.strpath):
-            url = _build_url(port, None, None)
-            cmd = "setup.py -vvv %s upload -r %s" % (pkg_frmt, url)
-            for i in range(5):
-                print('++Attempt #%s' % i)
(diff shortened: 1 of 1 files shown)


mplanchard:
I'm not sure supplying a default password is the right way to solve this. I started looking into this last week, but haven't found the source of the issue. The test that is hanging should not be requesting a password, as it's set to drop all auth.

dpkp:
It was caused by a change in setuptools. See:

https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v20100
https://github.com/pypa/setuptools/pull/555

dpkp:
I updated the username and password to be empty strings, which I believe is what you're attempting to test (no password required).

The other "fix" would be to attempt to set an older setuptools version in your tox deps configuration.

mplanchard:
Awesome. I thought it might be something in setuptools, but I have been quite busy and unable to look into in depth yet. Thanks for figuring it out! I'll try to get this reviewed sometime today, along with your other PRs.

mplanchard:
Looks good. Merging.

ankostis:
Thanks @dpkp for this fix, you really nailed it!

PR #146, opened 2016-06-23
Added docs for process management

mplanchard:
Added docs for process management systems `systemd` and `supervisor`


mplanchard:
Merging, as this is just docs.

PR #152, opened 2016-06-24
Docs, try/catch for PAM auth

mplanchard:
Updated some `rst` formatting in the docs, bumped the beta version, and removed the try/catch.

### DIFF
--- a/pypiserver/core.py
+++ b/pypiserver/core.py
@@ -92,12 +92,9 @@ def auth_by_htpasswd_file(htPsswdFile, username, password):
 
 
 def auth_by_pam(username, password):
-    try:
-        import pam
-        return pam.authenticate(username, password)
-    except ImportError as error:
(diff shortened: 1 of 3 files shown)


pypiserver/core.py
```diff
@@ -92,12 +92,9 @@ def auth_by_htpasswd_file(htPsswdFile, username, password):
...
 def auth_by_pam(username, password):
-    try:
-        import pam
-        return pam.authenticate(username, password)
-    except ImportError as error:
-        log.error('PAM module not found. Please install pam module')
-        return False
+    """Attempt to perform PAM auth. API access only for now."""
+    import pam
+    return pam.authenticate(username, password)
```
ankostis: This whole function is redundant - just use ` pypiserver.default_config(auther=pam.authenticate`, ...) directly.

mplanchard: Makes sense!

PR #153, opened 2016-06-24
Removed redundant function, added auther docs

mplanchard:
Also made some RST changes to the docs in general.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,11 +96,13 @@ Currently only password-protected uploads are supported!
    .. Tip:: When accessing pypiserver via the api, alternate authentication
       methods are available via the ``auther`` config flag. Any callable
       returning a boolean can be passed through to the pypiserver config in
-      order to provide custom authentication. For example, to authenticate
-      using the `pam module`_::
+      order to provide custom authentication. For example, to configure
+      pypiserver to authenticate using the `python-pam`_::
 
         import pam
         pypiserver.default_config(auther=pam.authenticate)
+
+      Please see `Running via the API`_ for more information.
 
 #. You  need to restart the server with the ``-P`` option only once
    (but user/password pairs can later be added or updated on the fly)::
@@ -500,6 +502,29 @@ management. An example configuration file for ``supervisor`` is given below::
 From there, the process can be managed via ``supervisord`` using ``supervisorctl``.
 
 
+Running via the API
+~~~~~~~~~~~~~~~~~~~
+
+Pypiserver can also be run via the API. The only additional feature that this
+enables is the use of alternative authentication mechanisms. The ``auther``
+keyword can be passed to the pypiserver app with any callable as a value. The
+associated callable should return a boolean when passed the username and
+password for a given request. Below is an example of starting the pypiserver
+from the Python shell, using ``/etc/passwd`` authentication via `python-pam`_::
+
+    >>> import pypiserver
+    >>> from pypiserver import bottle
+    >>> import pam
+    >>> app = pypiserver.app(port=80, auther=p.authenticate, root='/home/pypi/packages'
+    >>> bottle.run(app=app, host='0.0.0.0', port=80, server='auto')
+
+Bear in mind that whatever your authentication mechanism, the instance of
+Python running pypiserver must be able to access the appropriate resources in
+order for it to work. For example, the `python-pam`_ module, which enables
+authentication based on ``/etc/passwd``, seems to require root access in
+order to properly authenticate.
+
+
 Using a different WSGI server
 -----------------------------
 - *pypiserver* ships with it's own copy of bottle.
@@ -673,7 +698,7 @@ See the ``LICENSE.txt`` file.
 .. _PyPI: http://pypi.python.org
 .. _twine: https://pypi.python.org/pypi/twine
 .. _pypi-uploader: https://pypi.python.org/pypi/pypi-uploader
-.. _pam module: https://pypi.python.org/pypi/pam
+.. _python-pam: https://pypi.python.org/pypi/python-pam/
 .. |travis-status| image:: https://travis-ci.org/pypiserver/pypiserver.svg
     :alt: Travis build status
     :scale: 100%
--- a/README.rst
+++ b/README.rst
@@ -68,8 +68,8 @@ you may also upload them remotely with a ``python setup.py upload`` command.
 Currently only password-protected uploads are supported!
 
 #. First make sure you have the *passlib* module installed (note that
-   `passlib>=1.6` is required), which is needed for parsing the Apache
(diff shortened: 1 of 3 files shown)


mplanchard:
@ankostis will you have time to release 1.2.0 this weekend? I will be on vacation from Saturday through the 5th of July.

ankostis:
Yes

mplanchard:
Excellent. Any objections to me going ahead and merging this in?

ankostis:
No problem - it is just that i cannot reproduce it is working :-(

mplanchard:
What's not working? I did have to re-run the Travis job, because of a random failure not related to the code.

ankostis:
No no, i cannot auth using `/etc/passwd` !

mplanchard:
I'm having trouble getting the `pam` module to work at all on my vagrant box. The `pypiserver` code looks like it shouldn't have any problem, but I've never used the `pam` module.

mplanchard:
Are you running pypiserver as root? It seems like the user running python (and importing `pam`) needs to have root permissions in order for the `pam` authentication to work:

```
(pypi-venv) vagrant@vagrant-ubuntu-wily-64:~$ sudo /home/pypi/pypi-venv/bin/python
Python 3.4.3+ (default, Oct 14 2015, 16:03:50)
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pam
>>> p = pam.pam()
>>> p.authenticate('testuser', 'test')
True
>>> exit()
(pypi-venv) vagrant@vagrant-ubuntu-wily-64:~$ sudo su - pypi
pypi@vagrant-ubuntu-wily-64:~$ . pypi-venv/bin/activate
(pypi-venv) $ python
Python 3.4.3+ (default, Oct 14 2015, 16:03:50)
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pam
>>> p = pam.pam()
>>> p.authenticate('testuser', 'test')
False
````

**Comment by mplanchard** [comment]
Okay! Got it working. See below: 

On my vagrant box, I made a Linux user called `testuser` with the password `test`, then started the app as follows:

```
(pypi-venv) vagrant@vagrant-ubuntu-wily-64:~$ sudo /home/pypi/pypi-venv/bin/python
Python 3.4.3+ (default, Oct 14 2015, 16:03:50)
[GCC 5.2.1 20151010] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pypiserver
>>> from pypiserver import bottle
>>> import pam
>>> p = pam.pam()
>>>app = pypiserver.app(port=80, auther=p.authenticate, root='/home/pypi/packages', authenticated=['update','download'])

Bottle v0.13-dev server starting up (using AutoServer())...
Listening on http://0.0.0.0:80/
Hit Ctrl-C to quit.

10.0.2.2 - - [24/Jun/2016 19:44:25] "GET /requests/ HTTP/1.1" 303 0
10.0.2.2 - - [24/Jun/2016 19:44:25] "GET /simple/requests/ HTTP/1.1" 200 331
10.0.2.2 - - [24/Jun/2016 19:44:25] "GET /packages/requests-2.10.0-py2.py3-none-any.whl HTTP/1.1" 401 753
10.0.2.2 - - [24/Jun/2016 19:44:29] "GET /packages/requests-2.10.0-py2.py3-none-any.whl HTTP/1.1" 200 506877
```

Then from my computer:

```
(venv) [Nautilus@Nautilus-2 tmp]$ pip install --index http://localhost:8080 --trusted-host localhost requests
Collecting requests
User for localhost:8080: testuser
Password:
  Downloading http://localhost:8080/packages/requests-2.10.0-py2.py3-none-any.whl (506kB)
    100% |████████████████████████████████| 507kB 33.9MB/s
Installing collected packages: requests
Successfully installed requests-2.10.0
```

**Comment by mplanchard** [comment]
Seems to be working for uploads, too. Using this pypirc, I am able to upload successfully:

```
 23 [local-test]
 24 repository = http://127.0.0.1:8080
 25 username = testuser
 26 password = test
```

```
(venv) [Nautilus@Nautilus-2 dist]$ twine upload -r local-test pyupdate-0.1.0-py2.py3-none-any.whl
Uploading distributions to http://127.0.0.1:8080
Uploading pyupdate-0.1.0-py2.py3-none-any.whl
```

And when I change the password:
```
 23 [local-test]
 24 repository = http://127.0.0.1:8080
 25 username = testuser
 26 password = testers
```
I get an authentication error:
```
(venv) [Nautilus@Nautilus-2 dist]$ twine upload -r local-test pyupdate-0.1.0-py2.py3-none-any.whl
Uploading distributions to http://127.0.0.1:8080
Uploading pyupdate-0.1.0-py2.py3-none-any.whl
HTTPError: 403 Client Error: Forbidden for url: http://127.0.0.1:8080/
```

mplanchard:
I'll update the docs to reflect my findings.

mplanchard:
@ankostis see updated documentation for examples

ankostis:
Thanks @mplanchard!

Just a minor correction:

> Are you running pypiserver as root? It seems like the user running python (and importing pam) needs to have root permissions in order for the pam authentication to work:

[this SO](http://stackoverflow.com/a/32425797/548792) explains that it's not *root* that is need root - just read-access to `/etc/shadow`.

Also I saw that you invoked `pam.pam()`, so you installed `python_pam`, a package release on 2014 from the link you provide in the README, (not just `pam`, an older package from 2009).
Thanks for peeking this up.

In general, I really like this extension - it should be enabled by default for uploads, no need for fiddling with `htpasswd` etc
What do you think??.

mplanchard:
Ah, good to know it's not just root access that's required. And yes, `pam` apparently does not support Python 3, so I had to use `python-pam`, which seems to be better supported anyway. 

I like the idea of using /etc/passswd authentication out of the box. My only concern is that people who are  security conscious and like to run pypiserver as a user with very limited access might not want that user to have /etc/shadow access. Our setup is like that; we connect to a higher port that doesn't require root permission and let use nginx as a reverse proxy, so the pypiserver user only needs to be able to access its own virtual environment and the package root.

ankostis:
Yes, I was thinking the same security concern.
So let's opt-it-in this feature on the release.

blade2005:
So I like the direction this is going :) I'd like to also add that from http://stackoverflow.com/questions/5286321/pam-authentication-in-python-without-root-privileges

One way to accomplish this without root is to have that user be in the shadow group allowing read only access to the shadow file. 

Also if using other authentication method other than unix (pam default login) such as ldap root/shadow access isn't required.

Also thank you guys for supporting this project.

PR #161, opened 2016-09-06
Added receipe for reverse proxy configuration

ITler:
Added new receipe section for reverse proxy configuration with example configuration for nginx


mplanchard:
Looks very similar to the configuration I use! Thank you for the contribution.

PR #162, opened 2016-09-23
Feature/auther via cmd option

sakurai-youhei:
Because I'd like to use pypiserver with own auth provider via `python -m pypiserver`, I had motivation to introduce new command line option named `--auther`. FYI: I also fixed small problems eventually brought to me when preparing this pull-request.

### DIFF
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -7,6 +7,7 @@
 from __future__ import print_function
 
 import getopt
+import importlib
 import logging
 import os
 import re
@@ -119,6 +120,12 @@ def usage():
       Add "Cache-Control: max-age=AGE, public" header to package downloads.
       Pip 6+ needs this for caching.
 
+    --auther AUTHER
+      ad-hoc authentication provider to delegate authentication upon Python
+      callable object such as pam.authenticate provided by python-pam. As
+      pypi-server passes username and password, the callable shall return bool
+      value as an authentication result.
+
 
   pypi-server -h
   pypi-server --help
@@ -182,6 +189,7 @@ def main(argv=None):
             "log-err-frmt=",
             "welcome=",
             "cache-control=",
+            "auther=",
             "version",
             "help"
         ])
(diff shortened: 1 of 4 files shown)


pypiserver/__main__.py
```diff
@@ -235,7 +243,7 @@ def main(argv=None):
             c.password_file = v
         elif k in ("-o", "--overwrite"):
             c.overwrite = True
-        elif k in ("--hash-algo"):
+        elif k == "--hash-algo":
```
sakurai-youhei: The reason of this change is that `-h` was not working due to the fact that `("--hash-algo")` is equivalent to `"--hash-algo"` which contains `"-h"``.

ankostis:
Thank you for the fix, and apologies for the long response time.

pypiserver/__main__.py
```diff
@@ -249,6 +257,19 @@ def main(argv=None):
             c.log_err_frmt = v
         elif k == "--cache-control":
             c.cache_control = v
+        elif k == "--auther":
+            try:
+                mod, _, func = v.rpartition(".")
+                if mod:
+                    c.auther = getattr(importlib.import_module(mod), func)
```
ankostis: Using importlib is an indicator that you need an extension mechanism; *setuptools8 provide an excellent such mechanism through the `entry_points`.   
I prefer to leav this commit to wait until we introduce "plugins."

petri:
I need ability for more granular authentication: basically ability to have read-only users and also read+write (upload) users. That afaik is not possible currently.

Can we start an issue to design the first version of an authentication plugin that would allow to satisfy what @sakurai-youhei needs and also support more granular user-permission handling? I think using ABCs and matching entrypoints (by name) would work well. Or is the basic plugin implementation mechanism something that still needs more thought?

mplanchard:
We have yet to decide on the plugin implementation details. An issue like
you describe might help us to work through those details, but especially
with the US holidays coming up and it being a busy time of year trying to
get things done prior to that, I cannot guarantee immediate action.

That being said, in the meantime there are other ways of doing what you
describe. For example, if you reverse proxy to pypiserver using nginx, you
can use differential password authentication on different server locations
and http methods, allowing you to handle authentication one layer up from
pypiserver as it were.

Issue #167, opened 2017-01-30
Multiple Directory Support Broken


josephtate:
Somewhere between 1.1.6 and 1.2.0, the multiple directory support broke. I have this config:

```
[composite:main]
use = egg:Paste#urlmap
/pypi/unstable/ = unstable
/pypi/devel/ = devel
/pypi/stable/ = stable
/pypi/production/ = production

[app:unstable]
use = egg:pypiserver#main
root = /var/pypiserver/packages/unstable
password_file = /var/pypiserver/.htpasswd

[app:devel]
use = egg:pypiserver#main
root = /var/pypiserver/packages/devel
password_file = /var/pypiserver/.htpasswd

[app:stable]
use = egg:pypiserver#main
root = /var/pypiserver/packages/stable
password_file = /var/pypiserver/.htpasswd

[app:production]
use = egg:pypiserver#main
root = /var/pypiserver/packages/production
password_file = /var/pypiserver/.htpasswd

[server:main]
use = egg:gunicorn#main
host = 127.0.0.1
port = 8002
workers = 5
accesslog = -
```

Which works fine for 1.1.6-1.1.8, but under 1.2.0, each of the configured package directories show the same list of packages. I didn't narrow down WHICH directory is being served, but it's either the devel or unstable one (first or last in C ordering).

I could not run 1.1.9 and 1.1.10 successfully because of a paste_app_factory error.

mplanchard:
Thanks for the bug report @josephtate. We'll be looking into this. In the meantime, any extra information you could provide would be helpful including:

* operating system
* log output
* which of the directories is being served

Thanks!

josephtate:
The OS is Precise (Ubuntu 12.04) running python 2.7.3
The log output didn't have anything telling in it.
I think it's serving the first directory it finds (devel).

I don't have the setup anymore unfortunately, and it was running in production, so I can't take it down again to run more tests. I've locked it to version 1.1.8 in the meantime.

bertjwregeer:
https://github.com/pypiserver/pypiserver/blob/v1.2.0/pypiserver/__init__.py#L131

and 

https://github.com/pypiserver/pypiserver/blob/1.1.8/pypiserver/__init__.py#L42

Are vastly different.

The newer code imports `app` from `_app`, and returns it, however previously it would remove it from `sys.modules`, and re-import it, thereby creating a new `app`.

josephtate:
The pull request fixes the issue! Thanks @bertjwregeer

PR #173, opened 2017-03-31
Allow creating multiple servers using Paste

bertjwregeer:
This brings back the import hack that allows one to use Paste urlmap to
compose multipe PyPi servers within a single WSGI server.

After importing _app, we remove it from sys.modules so that future
imports will have their own copy, this allows globals to function
correctly and each _app to have their own variables.

Closes: https://github.com/pypiserver/pypiserver/issues/167

### DIFF
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -1,4 +1,5 @@
 import re as _re
+import sys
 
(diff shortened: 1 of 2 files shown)


bertjwregeer:
I've added a test that will make sure this regression doesn't happen again.

PR #190, opened 2017-11-25
Add $scheme forwarding to example nginx configuration

axnsan12:
Adding `X-Forwarded-Proto` makes bottle pick up on the original request's scheme (https/http), so that the URL on the welcome page is displayed correctly.


ankostis:
@axnsan12 thank you for the tip and the nice commit.
i take it as it is, assuming it works - not using nginx to test it.

Issue #163, opened 2016-09-24
Provide and extension mechanism based on setuptools entry_points


ankostis:
We are at the point where various PRs propose useful modifications that will begate the minimality of this project. (e.g   #162 for *auther*, #99 for site pages imitating *pypi*).
The typical way to satisfy all requirements is to allow them to grow separately from the project, based on a pluygin-extension mechanism.  That is exactly the role of [*setuptools* `entry_points`](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins).

The purpose is to enter

    pip install pypiserver[pypi_pages] 

and get a different installation to your requirements.

Implementation wise, the capabilities of the plugins must have been decided by the main app, and usually are done through the use of hook-lists and/or abstract base classes for configuring the app and the plugin(s). So we have to collect any "won't fix" issues candidates for pluginization, to establish the minimum requirements for the extension points.  The new label plugin `tag.plugins` should help to that.

petri:
+1 for defining interfaces using ABCs and matching (by name) entry points.

ankostis:
This simple guide on plugins from builtin python docs had missed me: https://packaging.python.org/guides/creating-and-discovering-plugins/

PR #99, opened 2015-12-21
Imitate PyPi's web-pages for packages

fraoustin:
add functionality a web page with url http://myserver/pypi.

Issue #93, opened 2015-10-12
Add functionnality http://myserver/pypi


fraoustin:
I created a fork on https://github.com/fraoustin/pypiserver for add functionality a web page with url http://myserver/pypi.

![pypiserver1](https://cloud.githubusercontent.com/assets/3457495/10423476/1dadffe8-70c7-11e5-8b33-ad293e21f8f2.png)
![pypiserver2](https://cloud.githubusercontent.com/assets/3457495/10423475/1dab3b6e-70c7-11e5-9f8e-1d18f79bb02c.png)

Are you interested?

PS: For full functionality , my fork needs the pkginfo module and docutils module

fraoustin:
add manage template in project  https://github.com/fraoustin/pypiweb

ankostis:
Dear @fraoustin would you mind making a regular pull-request?

ankostis:
Noting that this functionality had been rejected in #1, for its complexity.
At least this time we have to ensure it works flawlessly with TCs.

mplanchard:
@fraoustin, are there any packages at this point whose interdependence with pypiserver is harmed by the absence of this endpoint, such as `pyg`, referenced in #1?

ankostis:
Actually the *pypi* one day(?) will migrate to ["Warehouse"](https://pypi.org/) codebase and its pages will look like this:
![image](https://cloud.githubusercontent.com/assets/501585/17564541/f8cb48aa-5f33-11e6-937b-6d95338a5acd.png)
My point is that the project-pages better be customizable.

I suggest to use *setuptools* [entry-points](https://packaging.python.org/distributing/#entry-points) is a perfect match for converting this PR into the 1st plugin of the project.
@fraoustin, @mplanchard what do you think?

pawamoy:
IIUC, the goal here was to add a browsable web interface to pypiserver. Instead of baking it in, I'd suggest using simple-repository-browser, which looks well done: https://github.com/simple-repository/simple-repository-browser.

PR #197, opened 2017-12-03
WIP: Support setuptools entry-point plugins

ankostis:
In response to #163, implement an extension mechanism based on [setuptools entry-point](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins>) plugins. 

Note that it is based on `1.3.x` branch.

## TODO:
Stuff before removing *WIP* prefix from this PR:
- [ ] Although the the plugins mechanism is complete, actual plugins will have hard time (ie monkeypatch) to extend the app since extension "hooks" are missing (e.g. not possible to extend command-line options without monkey-patching `main()`).
- [ ] TestCases.
- [ ] Sources for a demo plugin would also be nice to have (e.g. convert #99 & #132 into the new architecture).

@petri check also this out.


pypiserver/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+Any collected ``<plugin-install-func>`` callables are invoked AFTER all
+plugin-modules have finished loading.
+
+.. Warning::
+   When appending into "hook" lists during installation, remember to avoid
+   re-inserting duplicate items.  In general try to well-behave even when
+   **plugins are initialized multiple times**!
+
+"""
+from _collections import OrderedDict
```
mplanchard: What's the reason for using the `_collections` extension module rather than `collections`?

ankostis: Oups! thanks

pypiserver/plugin.py
```diff
@@ -0,0 +1,88 @@
...
+_PLUGIN_GROUP_NAME = 'pypiserver.plugins'
+
+log = logging.getLogger(__name__)
+
+def _init_plugins(plugin_group_name=_PLUGIN_GROUP_NAME):
+    "Discover and load *setup-tools* plugins. "
+    global _plugins_installed
+
+    def stringify_EntryPoint(ep):
+        return "%r@%s" % (ep, ep.dist)
```
mplanchard: I'm not certain whether it would or wouldn't cause issues as it is, but since we're using a string literal here, we may want to add a `from __future__ import unicode_literals` import so that the keys of `_plugins_installed` will consistently be unicode string objects in both Python 2 and 3.

ankostis: Yeap.

ankostis: Will try both in travis and see the results.

mplanchard:
I like it! And I agree entirely with your TODO list. 

If you haven't already started on it, I can handle the conversion of command line argument parsing to `argparse`. 

Regarding hooks, we might want to compile (using the `tags.plugins` tagged issues) a list of places that *know* will need to check for implementations in the `_plugins_installed` dict. Defining `authers` is definitely one, for sure, and probably one of the most requested. It also seems like something like #93 could be relatively easy to implement if we allow additions to or overrides of the `bottle` route list.

Once we've got that, do we want to consider separating out different plugin types into separate groups, or do we want to maintain them all in the `plugins` group? A `routes` plugin group, for example, would indicate that everything loaded for that group should be added to the route list, and would keep us from having to do iteration and string searching or something similar through all defined plugins to determine which ones are meant to be loaded as routes. Something similar could potentially be done with `authers`.

ankostis:
> If you haven't already started on it, I can handle the conversion of command line argument parsing to argparse.

Be my guest!

> ... could be relatively easy to implement if we allow additions to or overrides of the bottle route list.

Actually we *cannot* disallow anything :-) 
The plugin suggested architecture here is quite permissive.  Plugin-authors don't have to abide to any contract (abstract class, configs, etc).  they can do whatever they wish on import time or in the init-function.  
Usually there they will have to monkey-patch module methods, and if they are lucky enough, for some things it would just suffice to augment/modify module-data (lists, sets, string-messages).

A stricter mindset (or more security-paranoid one) could dictate further down the road to define a class contract.

Anyhow, i'm preparing now 2 sample plugins for TCs to check yourself.

ankostis:
Let's hope we we're not gonna have problem with #173.. @mplanchard can you augment the plugin-tests, to check that?

Note: it was impossible to `pip install/uninstall` with python code to check them, because the same interpreter could not see the installed project immediately (but it was visible to the *next* TC.  Eventatually invoked shell-script from python with `subprocess` :-(

mplanchard:
I certainly can. I don't have any issue with testing the general plugin architecture via a `subprocess` script. We might be able to do something fancy with `tox` to get it t o install the plugins prior to running tests for the main package. I'll play around with it a bit and see if I can come up with something

ankostis:
Great thanks.

Regarding *paste* (#173), I'm incline to retrofit the existing hack as a plugin, which will be installed with `pip install pypiserver[paste]`, providing also with the `paste_factory()` code.
Does that make sense?

Regarding the TCs, i'm pasting here the `test_plugin.py` code that it DIDN'T play well in my case:
```
#! /usr/bin/env py.test

import logging

import pip
import pytest

import os.path as osp
import subprocess as sbp


mydir = osp.dirname(__file__)

installable_fpath = osp.join(tempfile.gettempdir(), 'installable.txt')
loadable_fpath = osp.join(tempfile.gettempdir(), 'loadable.txt')

def install_plugins():
    os.chdir(osp.join(osp.dirname(__file__), 'plugins'))
    pip.main('install -e .'.split())


def uninstall_plugins():
    try:
        pip.main('uninstall -y pypiserver-test-plugins'.split())
    except Exception as ex:
        logging.getLogger(__name__).warning(
            "Uninstalling `pypiserver-test-plugins` failed due to: %s!\n"
            "Probably never installed yet??" % ex)

def clean_mark_files():
    for fpath in [loadable_fpath, installable_fpath]:
        try:
            os.remove(fpath)
        except Exception as ex:
            logging.getLogger(__name__).warning(
                "Could not delete mark-file %s due to: %s!\n"
                "Probably this is the 1st run??" % (fpath, ex))


@pytest.fixture
def plugins_installed():
    clean_mark_files()
    install_plugins()
    try:
        yield
    finally:
        uninstall_plugins()
        clean_mark_files()


def test_no_plugins1():
    assert not osp.isfile(loadable_fpath)
    assert not osp.isfile(installable_fpath)

def test_plugins_smoketest(plugins_installed):
    assert not osp.isfile(loadable_fpath)
    assert not osp.isfile(installable_fpath)

    sys.modules.pop('pypiserver', None)
    import pypiserver

    assert osp.isfile(loadable_fpath)
    assert osp.isfile(installable_fpath)

def test_no_plugins2():
    assert not osp.isfile(loadable_fpath)
    assert not osp.isfile(installable_fpath)

```

mplanchard:
I think retrofitting the paste functionality as a plugin makes sense, but for the moment we still need to leave in the existing hack from #173, while throwing a `warning('This functionality has been moved to the pypiserver[paste] plugin and will be removed in a future relase', DeprecationWarning)`, to preserve backwards compatibility.

ankostis:
Agree.  We should add the deprecation warning for `1.x.x` releases.
And drop it in when a `2.x.x` release comes out.

ankostis:
OK, i found what was the problem with my all-python pip invocations in the `test_plugins.py` above: i was using `pip -e` (develop mode) and for some reason it did not update `sys.path` list with the new folder.
I can fix it either by appending the path in the script or by building and installing the package as *wheel*.

ankostis:
I still cannot *reliably* activate plugins in the same interpreter using pure python.
Discovered that on *linux* you need to invoke `pkg_resources.working_set.add_entry(package_dir)` after `pip install`.
I can now run it in *my* linux & Windows but not on travis! 
Have asked a more generic question there: https://stackoverflow.com/questions/47625482/write-test-cases-for-python-setuptools-entry-points-plugins

But in general I'm read to quit and rollback the last 2 experimental commits till a2084660.

petri:
What is pypiserver doing with the plugins it loads and (optionally) installs by calling some function in the plugin? Or if it's doing nothing, yet, what's the plan?

Quoting from above:

> Actually we cannot disallow anything :-)
> The plugin suggested architecture here is quite permissive. Plugin-authors don't have to abide to any contract (abstract class, configs, etc). they can do whatever they wish on import time or in the init-function.
> Usually there they will have to monkey-patch module methods, and if they are lucky enough, for some things it would just suffice to augment/modify module-data (lists, sets, string-messages).
> 
> A stricter mindset (or more security-paranoid one) could dictate further down the road to define a class contract.

I am not sure I understand how creating a (plugin) mechanism for monkey-patching provides much or any value? It is easy enough to create a package that monkey-patches another package. FWIW, I think there are quite a few packages out there on PyPI for helping with that as well.

What I am hoping for is that pypiserver would:

1. Define interfaces, using e.g. ABCs, or zope.interface, or just via documentation, for some parts of its functionality (such as getting list of valid authentication credentials), and entry point names that match those interfaces. For example:

```python
# The interface could be e.g.:

class AuthCredentialsProvider:
   "authentication credentials provider"

   def getAuthCredentials():
      "return the list of credentials as a list of (username, password) 2-tuples"

# Matching entry point name could be e.g. 'pypiserver:auth_credentials_provider', whatever
```

2. At startup, check whether there are plugins for the provided interfaces; if so, load (=import) them for later use

3. When e.g. authenticating, use the corresponding imported (e.g. auth) plugin

mplanchard:
@ankostis I'm totally fine with shell testing. I think it might be possible with `tox` to install a plugin package in the virtual environment prior to running the `tox` command, and have the plugin available as usual. That is, fundamentally, just shell testing, though, so I think either approach is good.

It's been busy at work this week, but I hope to have some time this weekend to finish up my argparse updates.

mplanchard:
@petri I don't think the two approaches are mutually exclusive.

The benefit of this approach is that it allows developers to define their interactions with pypiserver via the `entry_points` mechanism, which will allow pypiserver to automatically discover and load their patches/updates if they are installed. Otherwise, plugin authors would somehow have to make sure that users were loading their plugins at runtime, or running `pypiserver` via some kind of proxy offered by the installed plugin package. Please correct me if I'm wrong, obviously, but I think that's the primary advantage.

In the medium term, I also like the idea of adding plugin groups that function more directly in different interface areas.

ankostis:
The idea is to start low and simple for a couple of releases, and identify the extension points.
Then gradually retrofit these points to become extendable (eg `argparse` for cmd-line, authentication hooks ala-pam, ABCs for routes and behavior).

Regarding the monkey patching, I'm surprised that it can be achieved with pure package installations.
Is it reliable?
Wouldn't it be dependent on the order of the installations?
Could it support multiple "plugins" if each one require an incompatible strategy of bootstrapping, as mplanchard suggested?

Btw, "monkeypatching" is a rather misunderstood term in certain contexts. Overriding a class -method is a civilized and orderly way to monkeypatch.

petri:
I think we're talking apples and oranges here, coming from the opposite angles. If I've understood correctly, you are basically providing other developers means to have their code ("plugins") executed (via the entry points mechanism) when pypiserver is started, is that correct? So that they can dynamically modify pypiserver behavior as they deem necessary?

mplanchard:
Yes, that is exactly correct. It's essentially infinitely flexible, but a bit harder to developers to work with initially. However, we can get it  out quickly and then expand with more targeted interfaces from there.

PR #198, opened 2017-12-18
Drop support for EOL Python 3.3

### LINKED ISSUE #192: End Explicit Support for Python 3.3
Python 3.3 has been EoL'ed since September, and more and more major packages are dropping support, making maintenance more difficult. Pytest, for example, [dropped support in version 3.3.0](https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#pytest-330-2017-11-23).

I think that unless there is a really good reason, we should stop explicitly supporting Python 3.3 going forward.

hugovk:
Fix #192.

Also remove some redundant code for Python < 2.7 and 3.0-3.2, and some minor tidy-ups.

### DIFF
--- a/pypiserver/bottle.py
+++ b/pypiserver/bottle.py
@@ -1503,10 +1503,10 @@ class BaseResponse(object):
     # Header blacklist for specific response codes
     # (rfc2616 section 10.2.3 and 10.3.5)
     bad_headers = {
-        204: set(('Content-Type', )),
-        304: set(('Allow', 'Content-Encoding', 'Content-Language',
-                  'Content-Length', 'Content-Range', 'Content-Type',
-                  'Content-Md5', 'Last-Modified'))
+        204: {'Content-Type'},
+        304: {'Allow', 'Content-Encoding', 'Content-Language',
+              'Content-Length', 'Content-Range', 'Content-Type', 'Content-Md5',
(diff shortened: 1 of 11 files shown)


ankostis:
great!

pypiserver/bottle.py
```diff
@@ -89,17 +89,15 @@ def _cli_patch(args):
 
             def json_dumps(data):
                 raise ImportError(
-                    "JSON support requires Python 2.6 or simplejson.")
+                    "JSON support requires simplejson.")
```
ankostis: I would just not touch the `bottle.py` - that's an "upstream-derived" file for *pypiserver*, we don't maintain it.

I know that your changes are not invasive and make sanse, but 
there is another reason to keep this file pristine: 
when we upgrade *bottle*, [sometimes](https://github.com/pypiserver/pypiserver/commit/2d3f7bf) certain changes are needed to fix things.  Leaving `bottle.py` untouched makes these cases stand out, and be easier to redo those changes later, on each new upgrade.

hugovk: Makes sense, I'll revert those.

hugovk:
Rebased to resolve conflicts.

Note: latest master is also now failing: https://travis-ci.org/hugovk/pypiserver/builds/376047005

PR #239, opened 2019-01-30
Made the image smaller

ronneke1996:
Fixed bcrypt
Added requirements.txt


mplanchard:
This is great! Thank you!

I verified your updates locally, but I wanted to be sure that we reproduced this particular bug with an automated test, since I hadn't previously been able to reproduce it (my problem was my htpasswd files were using md5 and not bcrypt).

I opened a PR against your branch with some updates to the test file and other minor things. Once that's in, we can get this merged.

ronneke1996:
I've made a minor change to your test script and the travis config, to make the ci work.

mplanchard:
No, thank you! Glad we're finally putting this bcrypt issue to bed.

Issue #155, opened 2016-06-29
option to specify application base URL


luismsgomes:
Started pypiserver on port 127.0.0.1:8001 as follows:

```
$HOME/.pyenv/shims/pypi-server \
  --interface 12.0.0.1 --port 8001 \
  --authenticate update,download \
  --passwords $HOME/pypiserver/htpasswd.txt \
  --log-file $HOME/pypiserver/log.txt -vv \
  $HOME/pypiserver/packages
```

Then tried to reverse proxy the application through apache as follows:

```
ProxyPass  "/pypi/"  "http://127.0.0.1:8001/"
```

I can access the welcome page at the front-facing address `/pypi/` but the URLs in that page are wrong:
`/simple` should be `/pypi/simple`

Is it possible to add an option to `pypi-server --app-root=/pypi/` so that all URLs are prefixed with the given URL?

ankostis:
Unfortunately this is not as simple as it sounds since all urls are affected (see also #100 & #123).
Maybe a better configuration is to use `mod_wsgi` as described [in the document](/pypiserver/pypiserver#apache-mod-wsgi).
That way you get also multi-threaded responses for free.

Please provide some feedback if you attemt to setup it up.

PR #248, opened 2019-04-16
support changing the prefix of the path of the url

kujyp:
- Improvement of a known limitation.
The limitation is `It does not support changing the prefix of the path of the url (see #155 for workarounds).`
(https://github.com/pypiserver/pypiserver#known-limitations)
- Ref: https://github.com/pypiserver/pypiserver/issues/155
- Changed the links populated by `request.fullpath`. Now it uses `request.url`.

### Descriptions
There is an issue behind proxy using nginx with the prefix of the url.

In your code some links are populated with a proxy header `X-Forwarded-Host`.
Here's an example. `request.url` is changed by a proxy header `X-Forwarded-Host`.
https://github.com/pypiserver/pypiserver/blob/537034eb4323f0a84e66cb8d73e3b152c9d08744/pypiserver/_app.py#L92-L93

<br/><br/><br/><br/>

But `request.fullpath` isn't changed by a proxy header.
https://github.com/pypiserver/pypiserver/blob/537034eb4323f0a84e66cb8d73e3b152c9d08744/pypiserver/_app.py#L83-L96
fullpath
<br/><br/><br/><br/>

### Tests
I've tested with this nginx.conf file.
```
...
        location /pypi/ {
            proxy_set_header  X-Forwarded-Host $host:$server_port/pypi;
            proxy_set_header  X-Forwarded-Proto $scheme;
            proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header  X-Real-IP $remote_addr;
            proxy_pass        http://0.0.0.0:8001/;
        }
...
```

Full nginx.conf file: https://github.com/kujyp/pypiserver/blob/71ad25ee46f1afdf430d0f038c7a511bd2d3e3f3/proxytest/nginx.conf

I've tested with docker environments.
Here's the test assets for environments. https://github.com/kujyp/pypiserver/commit/71ad25ee46f1afdf430d0f038c7a511bd2d3e3f3
<br/>

I've tested with nginx, but not with other environments.
Let me know, if there are side effects on other environments.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -80,8 +80,6 @@ def favicon():
 
 @app.route('/')
 def root():
-    fp = request.fullpath
-
     try:
         numpkgs = len(list(packages()))
     except:
@@ -93,8 +91,8 @@ def root():
                     URL=request.url,
                     VERSION=__version__,
                     NUMPKGS=numpkgs,
-                    PACKAGES=urljoin(fp, "packages/"),
-                    SIMPLE=urljoin(fp, "simple/")
+                    PACKAGES=urljoin(request.url, "packages/"),
+                    SIMPLE=urljoin(request.url, "simple/")
                     )
 
 _bottle_upload_filename_re = re.compile(r'^[a-z0-9_.!+-]+$', re.I)
@@ -197,7 +195,7 @@ def update():
 @app.route('/packages')
 @auth("list")
 def pep_503_redirects(prefix=None):
-    return redirect(request.fullpath + "/", 301)
+    return redirect(request.url + "/", 301)
 
 
 @app.post('/RPC2')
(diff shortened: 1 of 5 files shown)


kujyp:
I added several test cases with `X_FORWARDED_HOST` header, too.

If this PR is against the intent of your code, let me know.

pypiserver/_app.py
```diff
@@ -59,6 +59,13 @@ def log_request():
     log.info(config.log_req_frmt, request.environ)
 
 
+@app.hook('before_request')
+def print_request():
+    parsed = urlparse(request.urlparts.scheme + "://" + request.urlparts.netloc)
+    request.custom_host = parsed.netloc
+    request.custom_fullpath = parsed.path.rstrip('/') + '/' + request.fullpath.lstrip('/')
```
kujyp: I used the names `request.custom_<property>` for additional properties to explicitly specify that these are not from `bottle.py`.

tests/test_core.py
```diff
@@ -96,7 +96,7 @@ def test_hashfile(tmpdir, algo, digest):
 def test_redirect_prefix_encodes_newlines():
     """Ensure raw newlines are url encoded in the generated redirect."""
     request = Namespace(
-        fullpath='/\nSet-Cookie:malicious=1;'
+        custom_fullpath='/\nSet-Cookie:malicious=1;'
```
kujyp: It should be changed, because it uses the `Namespace` instance not the `Bottle` instance.

skapin:
Well, perhapse i'm missing something but i got same behavior in recent build... How to fix it ?

PR #254, opened 2019-05-23
Do not force port 8080 in Dockerfile

wontonst:
I don't see a reason to force port 8080, I'd rather be able to do `docker run -p 80:8888 pypi --port 8888`

Personally my use case is when using AWS ECS under awsvpc networking mode, the container port is the host port so I need to be able to use `--port 80` to properly run the server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,5 +43,5 @@ USER pypiserver
 WORKDIR /data
 EXPOSE 8080
 
-ENTRYPOINT ["pypi-server", "-p", "8080"]
+ENTRYPOINT ["pypi-server"]
 CMD ["packages"]


mplanchard:
Are you not able to override the entrypoint for your task as described on [this page](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html)?

wontonst:
I could but it’s not a clean way to accomplish what I want.

mplanchard:
I'm sorry, but this is an easy configuration change to make in your local environment, and your suggested PR would break users' existing environments when they pulled the new image, which is unacceptable. Without updates to the docker-compose examples and README, it would also make our existing examples and documentation incorrect/broken.

Every container orchestration solution allows easy overriding of entrypoints and commands, so please use those solutions as appropriate for your individual needs or build your own Docker image with this one as a base.

PR #271, opened 2019-09-10
Correct a command-line argument parsing bug

maggyero:
This PR will correct a bug that affects the parsing of the `-h` command-line argument, which is the short version of the `--help` argument to print the help message.

Currently the command-line argument `-h` raises this error:

> $ pypi-server -h
> Error: while trying to list root(C:\Users\maggyero\packages): [WinError 3] Le chemin d’accès spécifié est introuvable: 'C:\\Users\\maggyero\\packages'

This is because the `-h` argument was intercepted by a preceding incorrect `elif` clause:

> elif k in "--hash-algo":  # incorrect, should be: elif k == "--hash-algo:"
> [...]
> elif k in ("-h", "--help"):

### DIFF
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -233,7 +233,7 @@ def main(argv=None):
             c.password_file = v
         elif k in ("-o", "--overwrite"):
             c.overwrite = True
-        elif k in "--hash-algo":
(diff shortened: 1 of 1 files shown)


maggyero:
@mplanchard

PR #276, opened 2019-10-16
Update welcome.html

maggyero:
Replace:

```
pip install --extra-index-url http://localhost:8080/ PACKAGE [PACKAGE2...]
```

with (adding the `simple/` path segment to the URI):

```
pip install --extra-index-url http://localhost:8080/simple/ PACKAGE [PACKAGE2...]
```

to avoid HTTP 303 redirections from http://localhost:8080/PACKAGE/ to http://localhost:8080/simple/PACKAGE/ by `pypiserver` when running the command to install PACKAGE.

### DIFF
--- a/pypiserver/welcome.html
+++ b/pypiserver/welcome.html
@@ -4,7 +4,7 @@
 
 <p> To use this server with pip, run the the following command:
 <blockquote><pre>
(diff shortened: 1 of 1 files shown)


mplanchard:
Looks good to me! Thanks for the contribution 🎉

PR #278, opened 2019-10-22
Update welcome.html

maggyero:
This PR will:

* replace the `--extra-index-url` option with `index-url` for the `pip install` command, since `pypi-server` redirects to PyPI anyway when needed;
* replace the short form `-i` option with the full form `--index-url` for the `easy_install` command, to be explicit and consistent with the `pip install` command;
* add a list of package arguments to the `easy_install` command to be consistent with the `pip install` command.


maggyero:
Hello @mplanchard, here is another update to the welcome page.

PR #284, opened 2019-11-12
Use python 3.6 for the Dockerfile

### LINKED ISSUE #281: Use Python 3 in Docker container
Hello,

I've noticed the Python version used to run pypiserver in its official Docker container is still 2.7.
Since it reaches EOL soon, would you object to upgrading to Python 3 ?

I'd like to work on this and submit a PR if you're interested. I might base the image on an official Python alpine image like `3.6-alpine`, if you're OK with that.

etene:
Closes #281

The resulting image should work the same way as the previous Python 2 based one.

I based the Docker image on `python:3.6-alpine` because I believe it is cleaner to use the official Python image for dockerized Python apps, however it comes at the cost of a larger image (~100M). Tell me if it's a problem and I'll find a fix.

I do not think it is possible to be as small as the Python 2 based image though, because Python seems to get slightly bigger at each release.

Using 3.7 or even 3.8 instead is just a matter of replacing the `FROM` statement, but I do not know if you support them yet so I played it safe.

Hope you will find this PR useful !

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-alpine as base
+FROM python:3.6-alpine3.10 as base
 
 # Copy the requirements & code and install them
 # Do this in a separate image in a separate directory
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,48 +1,33 @@
-FROM alpine:3.8 AS base
+FROM python:3.6-alpine as base
 
-# Install python and modules that can't be installed via pip
-# Delete the uncompiled variants to shave off ~10MB of the docker file
+# Copy the requirements & code and install them
+# Do this in a separate image in a separate directory
+# to not have all the build stuff in the final image
+FROM base AS builder
+RUN apk update
+# Needed to build cffi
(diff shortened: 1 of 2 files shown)


etene:
oh, looks like I broke something, I'll fix it

etene:
@mplanchard could you give this PR a look ? We're planning on deploying a pypiserver instance where I work, and I would really like to avoid deploying Python 2.7 based applications in late 2019 :)

heitorPB:
Do you think is a good idea to be more explicit about the alpine version used?
`FROM python:3.6-alpine as base` -> `FROM python:3.6-alpine3.10 as base` ?

etene:
@heitorPB That can't hurt, good idea !

etene:
The CI seems to have broken on its own... Let's try again.

etene:
An empty commit fixed it. @mplanchard any chance of this PR getting merged eventually ?

etene:
I totally understand, no problem, thanks for merging !

PR #294, opened 2020-01-12
DEP: Drop support for Py2 & Py3.4

mplanchard:
Both Python 2.x and Python 3.4 are EoL, so this drops official support for them. Pypiserver may continue to work with these versions, but they will no longer be tested. Bugs with EoL'ed versions will also no longer be fixed.

This also drops support for PyPy, while continuing to support PyPy3. This is because supporting PyPy implicitly means supporting Python 2.x, which I would prefer we not have to do.

This means the last version of pypiserver to support Python 2 and PyPy will be the 1.3.x series. From 1.4.0 forward, pypiserver will be officially supported for Python3 and PyPy3 only.


mplanchard:
Closing in favor of #321. We can drop py2 testing/support separately.

PR #295, opened 2020-01-19
Updated Docker Compose example with HTTPS configuration using Traefik

Lauszus:
Thanks for this project. I updated the Docker Compose example configuration, so it shows how to setup HTTPS using Traefik.


Lauszus:
@mplanchard you'r welcome :) Yes Traefik is great, it literally take get HTTPS up and running.

Btw which backend would you recommend? Right now it seems to take some time for it to respond to requests?

PR #298, opened 2020-02-24
Add the option to specify list of modules we don't want to update

peter-slovak:
When using `pypiserver` as both a partial mirror and a private package repository (e.g. company-owned codebase), regular package updates pose a security risk. For public packages, the updates work as expected - however, the updater also tries to locate the latest version of all the private packages.

Should an attacker find out a name of a private package, they could simply publish a higher version of said package into public `pypi.org`, containing arbitrary code (provided the package name is not being used publicly). Then, the `pypiserver` would pull the latest version at the time of the next update cycle.

This change adds the `--blacklist-file` option that is only considered along with the `-U` option (update). The argument value is a path to a file containing newline-separated package names that we want to skip when checking for latest package versions. This way, no version of a private package is ever downloaded from public PyPI if it's blacklisted.

P.S. Thank you for this great project :slightly_smiling_face:


pypiserver/core.py
```diff
@@ -222,7 +222,7 @@ def __init__(self, pkgname, version, fn=None, root=None, relfn=None, replaces=No
     def __repr__(self):
         return "%s(%s)" % (
             self.__class__.__name__,
-            ", ".join(["%s=%r" % (k, getattr(self, k))
+            ", ".join(["%s=%r" % (k, getattr(self, k, 'AttributeError'))
```
peter-slovak: Added this just to fix a `pytest` error when two `PkgFile`s didn't match - `pytest` was unable to show the diff, because artificially instantiated `PkgFile`s in tests didn't have the `_fname_and_hash` attribute. Instead of an object diff, I was getting:
```
tests/test_manage.py:216: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:919: in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:907: in assert_called_with
    raise AssertionError(_error_message()) from cause
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:901: in _error_message
    msg = self._format_mock_failure_message(args, kwargs)
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:799: in _format_mock_failure_message
    expected_string = self._format_mock_call_signature(args, kwargs)
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:794: in _format_mock_call_signature
    return _format_call_signature(name, args, kwargs)
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:2333: in _format_call_signature
    args_string = ', '.join([repr(arg) for arg in args])
../../../.pyenv/versions/3.8.0/lib/python3.8/unittest/mock.py:2333: in <listcomp>
    args_string = ', '.join([repr(arg) for arg in args])
pypiserver/core.py:225: in __repr__
    ", ".join(["%s=%r" % (k, getattr(self, k))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <list_iterator object at 0x7f2cfb03ab80>

>   ", ".join(["%s=%r" % (k, getattr(self, k))
                          for k in sorted(self.__slots__)]))
E                         AttributeError: _fname_and_hash
```

pypiserver/core.py
```diff
@@ -256,6 +256,22 @@ def _listdir(root):
...
+def read_lines(filename):
+    """
+    Read the contents of `filename`, stripping empty lines and '#'-comments.
+    Return a list of strings, containing the lines of the file.
+    """
+    lines = []
+
+    try:
+        with open(filename) as f:
+            lines = [line.strip() for line in f.readlines() if line.strip() and not line.strip().startswith('#')]
```
mplanchard, line 268: nit: we can avoid calling `strip()` so many times:

```suggestion
            lines = [
                line 
                for line in (ln.strip() for ln in f.readlines()) 
                if line and not line.startswith('#')
            ]
```

pypiserver/core.py
```diff
@@ -256,6 +256,22 @@ def _listdir(root):
...
+    except Exception:
+        pass
```
mplanchard, line 270: We may want to do a `log.warning` here or something, just so that there's some indication if the file failed to parse. I wonder if we might even want to exit on in the case of failure, if this is the sort of thing people are enabling for the sake of security

pypiserver/core.py
peter-slovak: :+1:

pypiserver/core.py
peter-slovak: That's probably a design / usability decision; personally I think there's too much space in `open()` to raise an exception and crash the whole pypiserver instance. Wrong file mode, privileges, path - just to name the most common ones. But you're absolutely right about the logging. I'll commit the change. 

If you or one of the maintainers decide that we don't want to start the server unless we read the blacklist file successfully, it should take just a `raise` statement to implement. I'm also more inclined to the security-before-availability approach, but don't want to take the burden of deciding this :slightly_smiling_face:

mplanchard: I totally understand!

I thought this through a bit today, and I think we should bomb if someone specifies a non-existent file or a file can't be read. That way, we can't get into a situation where someone thinks they're successfully preventing updates, but they aren't actually. If you want to go ahead and merge this, I can add the `raise` afterwards. Or feel free to add it if you're feeling like it.

Thanks again for the patch! We'll probably get a new version out here sometime in the next week.

peter-slovak: Sure, I added the `raise` with a bit of an explanation. Looking forward for the update! :slightly_smiling_face:

PR #300, opened 2020-03-17
feat: add download latest package route

davidt99:
Add download latest package route. I know it's not part of the pypi specification and pip does that automatically but I have an internal use case for that.
If the feature is accepted, I'll add tests and documentation.
Feel free to close it or ask any questions.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -369,6 +369,31 @@ def list_packages():
     return template(tmpl, links=links)
 
 
+@app.route("/packages/:prefix/latest")
+@auth("download")
+def download_latest_package(prefix):
+    # PEP 503: require normalized prefix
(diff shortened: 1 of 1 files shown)


dee-me-tree-or-love:
Hey @davidt99 ! Thanks a lot for your PR. First of all, sorry for the late notice on this change, it's been a long time getting into the project and besides I've had a lot of activities on all sides lately. But more related to this feature, I am very curious if you could elaborate if this is still a requirement for you and if you are still using this? 

I am asking because I know that there has been an idea to open up a plugin mechanism to the developers so that the base pypiserver can be extended and this sounds like a great candidate for this. So I'll be very happy to hear a little more about the context of this idea, to know better how to move forward. 

All the best ✌️

PR #316, opened 2020-06-05
Added running as a service in windows with NSSM to readme

kodaman2:
- Added detailed steps in how to run pypiserver as a service in Windows OS.

Quick and simple patch. I've tested in my local environment and it works great!

Server machine:
image

Installing packages from a remote machine:
image


kodaman2:
Glad to contribute!

PR #326, opened 2020-07-30
Moved flask-pypi-proxy and pip2pi to a new "Unmaintained or archived"…

Luttik:
Moved flask-pypi-proxy and pip2pi to a new "Unmaintained or archived" header since the former is archived and the latter has seen barely any development since mid 2015 and no development since sep 2019.


README.rst
```diff
@@ -951,16 +951,21 @@ among the most popular alternatives:
   <https://pypi.org/project/devpi/>`_.
   (version: 2.1.4, access date: 8/3/2015)
 
+- Check this SO question: `How to roll my own pypi
+  <http://stackoverflow.com/questions/1235331/how-to-roll-my-own-pypi>`_
```
dee-me-tree-or-love, line 955: Why do you think this is necessary? It is a little vague why would it be helpful to read this question. I would say either elaborate on why or remove.

README.rst
```diff
@@ -951,16 +951,21 @@ among the most popular alternatives:
...
+------------------------
+
+These projects were once alternatives to pypiserver but are now either unmaintained or archived.
+
 - `pip2pi <https://github.com/wolever/pip2pi>`_
   a simple cmd-line tool that builds a PyPI-compatible local folder from pip requirements
-  (version: 0.6.7, access date: 8/3/2015)
 
 - `flask-pypi-proxy <http://flask-pypi-proxy.readthedocs.org/>`_
   A proxy for PyPI that also enables also uploading custom packages.
```
dee-me-tree-or-love, line 967: :+1: good catch

README.rst
```diff
@@ -951,16 +951,21 @@ among the most popular alternatives:
...
+
+
+Unmaintained or archived
+------------------------
+
+These projects were once alternatives to pypiserver but are now either unmaintained or archived.
+
 - `pip2pi <https://github.com/wolever/pip2pi>`_
   a simple cmd-line tool that builds a PyPI-compatible local folder from pip requirements
-  (version: 0.6.7, access date: 8/3/2015)
```
dee-me-tree-or-love, line 956: Good that this version is removed but I am not sure why do you consider this project unmaintained, it is not very active indeed but still the last commit was about a year ago and the project is not archived. Maybe I have not found something in the issues or other places, could you correct me if I am wrong?

README.rst
```diff
@@ -980,7 +985,6 @@ may want to familiarize with:
   the software that powers PyPI_ itself. It is not generally intended to
   be run by end-users.
 
-
```
dee-me-tree-or-love, line 983: I am not a big expert on RST syntax and also prefer markdown myself, but, for consistency, other headings with `===` level also have two blank lines before them, maybe it's better to keep this one

dee-me-tree-or-love (commented):
Nice, it's good to keep the README up to date, thanks for checking this, could you see the comments? But on all other sides, I'd be glad to merge soon

README.rst
Luttik: Yeah, this totally depends on your definition of unmaintained. It has no commits in about a year, no release more than 1.5 years, and unanswered issues going back to 2015. That is unmaintained in my book, but I understand if you want to be more conservative in your readme. On the other hand, I wouldn't like to be the one who forwards others to something that is this stale.

README.rst
Luttik: Sure.

README.rst
dee-me-tree-or-love: Fair enough, that's indeed very subjective, you've got a point here. :ok_hand: don't mean to offend the owners of the other project but that's fair enough, they can poke us back :)

README.rst
mplanchard: FWIW pypi does support markdown now, provided you specify a config value in the setup.py, so it might be worth switching this to markdown at some point. I think very few people ever have preferred RST :)

mplanchard (approved):
:+1:

Issue #309, opened 2020-05-17
[BUG] Upload failed (500): with volume


Hedgehogues:
I start such configuration of docker-compose
```
  pypiserver:
      restart: $RESTART
      image: pypiserver/pypiserver:latest
      volumes:
          - ./.htpasswd:/data/auth/.htpasswd
          - ./packages:/data/packages
      command: -P /data/auth/.htpasswd -a update,download,list /data/packages
      ports:
          - "8081:8080"
```

I have a problem with volume of packages: ` - ./packages:/data/packages`. If I forgot about packages volume, it's all right. But with this line, I have next line while I upload my test package with next command:

`python setup.py bdist_wheel upload -r test`

I have next log:

```
['loguru==0.4.1', 'fastapi==0.54.1', 'uvicorn==0.11.5', 'starlette==0.13.2', 'vault-client==0.3.2', 'prometheus_client==0.7.1']
running bdist_wheel
running build
running build_py
installing to build/bdist.linux-x86_64/wheel
running install
running install_lib
creating build/bdist.linux-x86_64/wheel
creating build/bdist.linux-x86_64/wheel/xxx
copying build/lib/xxx/__init__.py -> build/bdist.linux-x86_64/wheel/xxx
running install_egg_info
running egg_info
writing xxx.egg-info/PKG-INFO
writing dependency_links to xxx.egg-info/dependency_links.txt
writing entry points to xxx.egg-info/entry_points.txt
writing requirements to xxx.egg-info/requires.txt
writing top-level names to xxx.egg-info/top_level.txt
reading manifest file 'xxx.egg-info/SOURCES.txt'
writing manifest file 'xxx.egg-info/SOURCES.txt'
Copying xxx.egg-info to build/bdist.linux-x86_64/wheel/xxx-0.1.0-py3.7.egg-info
running install_scripts
creating build/bdist.linux-x86_64/wheel/xxx-0.1.0.dist-info/WHEEL
creating 'dist/xxx-0.1.0-py3-none-any.whl' and adding 'build/bdist.linux-x86_64/wheel' to it
adding 'xxx/__init__.py'
adding 'xxx-0.1.0.dist-info/METADATA'
adding 'xxx-0.1.0.dist-info/WHEEL'
adding 'xxx-0.1.0.dist-info/entry_points.txt'
adding 'xxx-0.1.0.dist-info/top_level.txt'
adding 'xxx-0.1.0.dist-info/RECORD'
[...]

Issue #306, opened 2020-05-07
Permission for /data/auth


olefemmy:
The permission for authentication does not seem to work as expected, and this seems to be due to the fact that `/data/auth` is not owned by `pypiserver user`.

If what was done for `/data/packages` is done for `/data/auth`, then it works.

So, something like `mkdir -p /data/auth  && chmod g+s /data/auth` should work in your image.

mplanchard:
Are you running the container on linux?

olefemmy:
Yes, on Linux...

mplanchard:
The issue w/volumes in Linux is that the docker engine doesn't automatically run the container as the host user, so accessing things within the volume is problematic. I'm a little more hesitant to run `chmod` on users' auth stuff than on the package directory by default. I think it might be okay, but I'd like to think about it a bit more. In the meantime, I think the workarounds described [here](https://github.com/pypiserver/pypiserver/issues/256#issuecomment-552247106) should probably work for you.

olefemmy:
Oh no, I have fixed the problem locally like I suggested. I was only suggesting improvements.

mplanchard:
The docker version in the current master branch (published with the `unstable` tag) makes some updates to the way we handle permissions. We don't try to change the permissions on mounted volumes, but we are a little more flexible about access, so as long as you're mounting the `/data/auth` in such a way that the container user can read it, it should work, and if not, it should give you I think a better error message.

If you get the chance, could you pull `unstable` and let us know whether this addresses your issues, and if not, how you think we might be able to further improve the experience?

Issue #256, opened 2019-05-29
Permission issues when using existing packages folder with Docker setup


a-recknagel:
### How to reproduce
```bash
mdkir packages
docker run -d -p 8080:8080 -v ~/tmp/pypiserver/packages:/data/packages pypiserver/pypiserver:latest
twine upload --repository-url http://0.0.0.0:8080 my/local/package.whl
```
### Server logs
```
172.17.0.1 - - [29/May/2019 06:48:04] "POST / HTTP/1.1" 500 745
Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/pypiserver/bottle.py", line 923, in _handle
    return route.call(**args)
  File "build/bdist.linux-x86_64/egg/pypiserver/bottle.py", line 1845, in wrapper
    rv = callback(*a, **ka)
  File "build/bdist.linux-x86_64/egg/pypiserver/_app.py", line 52, in protector
    return method(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/pypiserver/_app.py", line 195, in update
    file_upload()
  File "build/bdist.linux-x86_64/egg/pypiserver/_app.py", line 172, in file_upload
    core.store(packages.root, uf.raw_filename, uf.save)
  File "build/bdist.linux-x86_64/egg/pypiserver/core.py", line 283, in store
    save_method(dest_fn, overwrite=True)  # Overwite check earlier.
  File "build/bdist.linux-x86_64/egg/pypiserver/bottle.py", line 2447, in save
    with open(destination, 'wb') as fp:
IOError: [Errno 13] Permission denied: '/data/packages/package.whl'
```
### Workaround

```bash
mdkir packages
chown -R 9898:98989 packages  # the internal user id can be looked up in the Dockerfile
docker run -d -p 8080:8080 -v ~/tmp/pypiserver/packages:/data/packages pypiserver/pypiserver:latest
twine upload --repository-url http://0.0.0.0:8080 my/local/package.whl
```
Doing this seems like a hack, and is not documented in [the section on docker deployments](https://pypi.org/project/pypiserver/#using-the-docker-image). Am I doing something wrong?

mplanchard:
Hm. I'm sorry you're getting this error!

I didn't have to do this locally when I was testing out the docker deployments. Could you give me a bit more information about the system you're running on?

Once we get the root cause figured out, I'll either update the docs to make things clearer or update the docker config to make it easier.

[...]

PR #330, opened 2020-09-01
Update Dockerfile. Update docker base images.

normoes:
This PR udpates the Dockerfile.

Docker base images are updated.

The main change is the introduction of an `entrypoint` script, which sets the permission on the package directry `/data/packages`.

In the previous version of the `Dockerfile` this was done before declaring the same directory as docker `VOLUME`, which switched ownership back to `root`.

This behaviour caused: https://github.com/pypiserver/pypiserver/issues/309

To still run `pypiserver` as user and not `root`, the tool `su-exec` was introduced and used within `entrypoint.sh`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,64 @@
-FROM python:3.6-alpine3.10 as base
+FROM python:3.8-alpine3.12 as base
 
 # Copy the requirements & code and install them
 # Do this in a separate image in a separate directory
 # to not have all the build stuff in the final image
-FROM base AS builder
-RUN apk update
-# Needed to build cffi
-RUN apk add python-dev build-base libffi-dev
+FROM base AS builder_gosu
+
+ENV GOSU_VERSION 1.12
+
+RUN apk add --no-cache --virtual .build-deps \
+        ca-certificates \
+        dpkg \
+        gnupg \
+    && dpkgArch="$(dpkg --print-architecture | awk -F- '{ print $NF }')" \
+    && wget -O /usr/local/bin/gosu https://github.com/tianon/gosu/releases/download/${GOSU_VERSION}/gosu-${dpkgArch} \
+    && wget -O /usr/local/bin/gosu.asc https://github.com/tianon/gosu/releases/download/${GOSU_VERSION}/gosu-${dpkgArch}.asc \
+    # verify the signature
+    && export GNUPGHOME="$(mktemp -d)" \
+    && gpg --batch --keyserver hkps://keys.openpgp.org --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4 \
+    && gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu \
+    && command -v gpgconf && gpgconf --kill all || true \
+    && rm -rf "$GNUPGHOME" /usr/local/bin/gosu.asc \
+    && chmod +x /usr/local/bin/gosu \
+    # check installation
+    && gosu --version \
+    && gosu nobody true \
+    && apk del --no-cache \
+        .build-deps \
+    && rm -rf /var/cache/apk/* \
+    && rm -rf /tmp/*
+
+FROM base AS builder_dependencies
+
 COPY . /code
-RUN mkdir /install
-RUN pip install --no-warn-script-location \
+
+RUN apk add --no-cache --virtual .build-deps \
+        build-base \
+        libffi-dev \
+    && mkdir /install \
+    && python -m pip install --no-warn-script-location \
                 --prefix=/install \
-                /code --requirement /code/docker-requirements.txt
+                /code --requirement /code/docker-requirements.txt \
+    && apk del --no-cache \
+        .build-deps \
+    && rm -rf /var/cache/apk/* \
+    && rm -rf /tmp/*
 
 FROM base
+# Copy the libraries installed via pip
+COPY --from=builder_dependencies /install /usr/local
+COPY --from=builder_gosu /usr/local/bin/gosu /usr/local/bin/gosu
+COPY entrypoint.sh /entrypoint.sh
 
 RUN addgroup -S -g 9898 pypiserver \
     && adduser -S -u 9898 -G pypiserver pypiserver \
     && mkdir -p /data/packages \
-    && chown -R pypiserver:pypiserver /data/packages \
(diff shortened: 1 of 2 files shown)


normoes:
https://github.com/pypiserver/pypiserver/issues/306

normoes:
https://github.com/pypiserver/pypiserver/issues/256

Dockerfile
```diff
@@ -1,33 +1,48 @@
-FROM python:3.6-alpine3.10 as base
+FROM python:3.7-alpine3.12 as base
```
mplanchard, line 1: let's go ahead and update this to 3.8 while we're at it
```suggestion
FROM python:3.8-alpine3.12 as base
```

Dockerfile
```diff
@@ -1,33 +1,48 @@
...
+RUN apk add --no-cache \
+        python3 \
+    && apk add --no-cache --virtual .build-deps \
+        git \
+        gcc \
+        build-base \
+        libffi-dev \
+    && mkdir /stuff && cd /stuff || exit 1 \
+    && git clone https://github.com/ncopa/su-exec.git su-exec-clone \
+    && cd su-exec-clone || exit 1 \
```
mplanchard, line 17: I'd prefer we use [gosu](https://github.com/tianon/gosu) instead of `su-exec`. I recognize it's larger, but they publish built binaries for their releases, which we could use rather than pulling in all the build machinery, cloning, making, etc. I'm also personally more familiar with `gosu`, and it seems to be the more commonly used of the two (also easier for most of us to introspect go dependencies than C)

entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/ash
+
+set -euo pipefail
+
+chown -R pypiserver:pypiserver /data
+# Set the setgid bit so anything added here gets associated with the
+# pypiserver group
+# chmod g+s /data/packages
```
mplanchard, line 8: Is this commented out intentionally?

entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ "$@" = "" ]; then
+    # default CMD
+    echo "Set default option '/data/packages'"
+    set -- " /data/packages"
+else
+    # 
+    echo "Using custom CMD: $@"
+fi
```
mplanchard, line 17: I'm a little concerned that this update could break people's existing commands. If someone were relying on `packages` being provided by default, and were adding extra commands on top of that, it seems like this might break, unless I'm missing something?

e.g. if someone were running `docker run pypiserver -- log-file log.out`?

mplanchard (commented):
I like that this adds the ability to set the port to a non-default value for folks who need that, and I'm glad we're hopefully resolving the permissions issues in a cross-platform way! I have a few questions and suggestions re: the implementation

entrypoint.sh
normoes: Right, I will remove those comments.

It's not necessary to set permissions like that, I found.

Dockerfile
normoes: Good idea.

Dockerfile
normoes: In general I agree, but this would also mean to:
* Install `Go` into the docker image or
* Use a `Go` based docker image and install `python`. 

This also means:
* The docker image gets bigger.
* There will be an additional system dependency.

This is basically why I decided against `gosu` in the first place.

Just a few notes to the current implementation (coming with this PR):
* The runtime docker image is based on the `base` image, which does not contain the build dependencies - it is pure `python:3.8-alpine3.12`.
* The `builder` image (which also builds `su-exec`) is cleaned at the very end and the build dependencies are removed as well.

If you like, I can still implement `gosu`.

entrypoint.sh
normoes: I tested a little bit. So, as far as I can tell, the above command `docker run pypiserver --log-file log.out` won't work in the current implementation, because the `packages` directory is not added by default as soon as docker gets additional parameters.
```
Error: while trying to list root(/home/pypiserver/packages): [Errno 2] No such file or directory: '/home/pypiserver/packages'
```
Ok, there is a default `packages` path: `/home/pypiserver/packages`, but it does not exist in the docker container.

---

The default `packages` path is added because of
```
CMD ["packages"]
```
However, this only works in case no additional docker commands are given when the container is started.
The only option, that's added every time is the port, because - in the original implementation - it is part of the `ENTRYPOINT`:
```
ENTRYPOINT ["pypi-server", "-p", "8080"]
```

---

The given command in the Dockerfile's `CMD` will be overwritten when doing
```
docker run pypiserver --log-file log.out
```

The command should be
```
docker run pypiserver --log-file log.out packages
```
to make it work properly.

---

I think the new implementation (contained in this PR) behaves in exactly the same way the original implementation does.

Of course it is possible to always add the default path, but I think, this will get confusing and can, in the end, actually affect user's commands.
What do you think?

Dockerfile
mplanchard: Do you need go to run a go binary? I thought they could run without golang installed on the system. `gosu` publishes binary releases, so I figured we might be able to use that.

If that's not the case, no need to bother with `gosu`!

entrypoint.sh
mplanchard: Thanks for digging in on that! I think you're right. Any specification of additional arguments on the current dockerfile would override the default `packages` argument, so I think this will behave exactly like it did before.

mplanchard (dismissed):
I'm good w/this going in either as-is or with `gosu`! Let me know what you find w/that

Dockerfile
normoes: I didn't know - I tested it and it works without installing `Go` itself. Great. Changed :+1: 

Some changes:
* I replaced `su-exec` with `gosu`.
* I added the signature check (for `gosu`) according to the recommendation from https://github.com/tianon/gosu/blob/master/INSTALL.md.
* I restructured the docker layers to separate the single build stages and their system dependencies. Also this separates the data within those intermediary docker images, even makes them smaller. - This only affects the docker image build.

Could you please have another look @mplanchard ?

Thanks a lot.

mplanchard (approved):
Thanks again for this!

mplanchard:
Also just noting that I've created a [zulip](https://pypiserver.zulipchat.com) to help organize community discussions, feature ideas, etc. Feel free to join if you'd like.

PR #334, opened 2020-09-09
Log messages to stdout instead of stderr

### LINKED ISSUE #328: access log printed to stderr
Hi,

I've noticed that access log messages like:
`[31/Jul/2020 18:13:23] "GET / HTTP/1.1" 200 1009`
are printed to stderr. Cloud monitoring systems like stackdriver qualifies these messages as errors, so as you may presume error reports list is polluted.

I've also noticed that there is no simple way to change this behavior. Docs mention `--log-conf` cli parameter, but it isn't used anywhere in code.

Am i missing something or is it a bug?
Cheers.

elfjes:
fixes #328


pypiserver/__main__.py
```diff
@@ -20,7 +20,7 @@
 
 
 def init_logging(level=None, frmt=None, filename=None):
-    logging.basicConfig(level=level, format=frmt)
+    logging.basicConfig(level=level, format=frmt, stream=sys.stdout)
```
dee-me-tree-or-love, line 23: Great, could you make it also optional and possible to specify from the configuration? Thanks!

pypiserver/__main__.py
```diff
@@ -117,6 +120,9 @@ def usage():
       A format-string selecting Http-Error properties to log; set to  '%s' to see them all.
       [Default: %(body)s: %(exception)s \n%(traceback)s]
 
+    --log-to-stdout
+      log messages to stdout instead of the default stderr
```
mplanchard, line 124: Since we're touching this, it might be better to give people some more configurability with this, for example:

```
--log-to-stream
  Log messages to the specified stream. Valid values are "stdout", "stderr", or "none"
```

That way people can turn off string logging altogether if they want. With your refactor to the `init_logging` function, this should be pretty easy to add in. What do you think?

mplanchard (commented):
Thanks for this patch! This was a good catch, and I like the direction. I have a few thoughts about ways we might be able to improve the logging experience a bit more. Let me know what you think!

pypiserver/__main__.py
elfjes: Sounds good. I've added it to the PR. I've renamed it to `log-stream` to be more consistent with the `log-file` option

elfjes:
It took me a while to notice that the pipeline was failing, but it is fixed now.

mplanchard (approved):
This looks good to me! Thanks very much for the contribution, and feel free to join us on [zulip](https://pypiserver.zulipchat.com)

PR #336, opened 2020-09-26
Run black on codebase

### LINKED ISSUE #335: Use black as a code formatter
After reading #325 I think implementing black as a code formatter is maybe the lowest hanging fruit here. I think a MR for this would consist of the following:
- [ ] determine the max line lenght, or just stick to the default
- [ ] run black on the whole code base
- [ ] run a travis job `black --check` to ensure all commits are properly formatted

What do you think of this idea?

elfjes:
fixes #335 
- [x] blacken all files
- [x] add black check stage to travis


mplanchard:
Thanks for this! I have a couple of nits:

- Could we add a `pyproject.toml` to configure black and set it to 80 character line lengths? I know I am probably in an ever growing minority of people who still prefer 80 characters, but I like being able to have two or three editor windows open side-by-side
- Could we also in that config file exclude `bottle.py` from the list of files to format? Since it's a vendored dependency, I'd like to keep it as close to the original as possible in case we need to pull in updates

elfjes:
I've picked your pyproject.toml and added a few more exclusion lines from the default black exclusion list (.tox and .direnv)

mplanchard (approved):
Thanks for pulling in that config!

PR #339, opened 2020-10-05
Backwards-compatible argparse config (not yet in use)

mplanchard:
Adds an argparse config that, while adding subcommands (`pypi-server run` and `pypi-server update`), retains full commandline backwards compatibility with the existing config parsing logic.

There's a bit of hackery required to do this, so this also issues a warning if using the non-subcommand arguments, allowing us to potentially remove support for the old form in our next next major version bump (i.e. 3.0).

Also adds a `.pyproject.toml` with a black config, and a mypy config block to `setup.cfg`.

`mypy` is now called in `tox`, currently only for `config.py`, because nothing else typechecks successfully.


pypiserver/config.py
```diff
@@ -0,0 +1,599 @@
...
+        if update_idx == -1:
+            # We were a "run" command.
+            args.insert(insertion_idx, "run")
+        else:
+            # Remove the -U from the args and add the "update" command at the
+            # start of the arg list.
+            args.pop(update_idx)
+            args.insert(insertion_idx, "update")
+
+        return args
```
elfjes, line 587: What happens when someone calls `pypiserver run -U`?

elfjes (commented):
Looks good :) I'm kind of surprised that it requires this amount of hacking to be backwards compatible, but it's a nice solution, considering what you had to work with.

When do you plan on integrating this?

pypiserver/config.py
```diff
@@ -0,0 +1,599 @@
...
+        raise ValueError(
+            f"Hash algorithm {arg} is not available. Please select one "
+            f"of {hashlib.algorithms_available}"
+        )
+    return getattr(hashlib, arg)
+
+
+def html_file_arg(arg: t.Optional[str]) -> str:
+    """Parse the provided HTML file and return its contents."""
+    if arg is None or arg == "pypiserver/welcome.html":
```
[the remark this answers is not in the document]
mplanchard: The only reason I didn't was because I'm currently using the formatter that automatically adds the defaults to the help text, and I didn't want it to look nasty. However, I think I need to switch to another formatter anyway so that I can get more granular formatting of some of the multiline helps with examples, so I'll make this change too

pypiserver/config.py
```diff
@@ -0,0 +1,599 @@
...
+        return pkg_resources.resource_string(__name__, "welcome.html").decode(
+            "utf-8"
+        )
+    with open(arg, "r", encoding="utf-8") as f:
```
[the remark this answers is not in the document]
mplanchard: I keep forgetting we're in Python3-land now :)

pypiserver/config.py
```diff
@@ -0,0 +1,599 @@
...
+                # No command was found, which happens if no arguments were
+                # provided. We will raise a SystemExit so we get the same
+                # deprecated argument handling functionality as we do for
+                # a failed parse.
+                orig_err = None
+                raise SystemExit
+        except SystemExit:
+            # argparse could not parse the arguments. Perhaps they are
+            # old-style arguments, so we'll try adjusting and re-parsing.
+            err_msg = (
```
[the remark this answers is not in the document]
mplanchard: Yeah, normally I'm a stickler for explicit null checks, to avoid unexpected bugs with falsy values, but I think we'd want to use the usage message if this is an empty string anyway, so the falsy check will be good

pypiserver/config.py
mplanchard: argparse would throw, because it parses the `run` subcommand first, and recognizes that `-U` is not a valid argument for that subcommand.

Adding this in to the bottom of this file:

```py
print(Config.from_args(["run", "-U"]))
```

I get:

```
usage: config.py [-h] {run,update} ...
config.py: error: unrecognized arguments: -U
```

mplanchard:
> Looks good :) I'm kind of surprised that it requires this amount of hacking to be backwards compatible, but it's a nice solution, considering what you had to work with.

Most of the hackiness is around trying to display nice messages, tbh. I'm going to spend a bit of time over the next few days cleaning this up a bit more, and then I'll get it merged.

> When do you plan on integrating this?

Trying to get this slotted in to replace the current config will be the next thing I'm working on... the activity in the modernization PR has given me a bit more motivation to work on this project than usual, so I'm hoping I can get that done this upcoming weekend :)

mplanchard:
Thanks very much for the review, by the way!

pypiserver/config.py
```diff
@@ -0,0 +1,664 @@
...
+        raise ValueError(
+            "Invalid authentication options. `.` (no authentication) "
+            "must be specified alone."
+        )
+    return items
+
+
+def hash_algo_arg(arg: str) -> t.Callable:
+    """Parse a hash algorithm from the string."""
+    if arg not in hashlib.algorithms_available:
```
elfjes, line 86: Question here: it looks like disabling the hash_algo doesnt work anymore, or am I missing something?

mplanchard: Ah, thanks for catching this. I forgot to handle the "falsey" cases. I'll add that in.

mplanchard: Added in #347

PR #342, opened 2020-10-07
Cleanup code to python 3.6

elfjes:
* Remove Python version conditional code
* Remove explicit inheritance from object
* Cleanup setup.py
* Convert most string interpolations to f-strings
* Convert some file io to use pathlib

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,8 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
-import sys
+from pathlib import Path
 
 from setuptools import setup
-
-if sys.version_info >= (3, 0):
-    exec("def do_exec(co, loc): exec(co, loc)\n")
-else:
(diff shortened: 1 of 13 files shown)


setup.py
```diff
@@ -17,33 +12,31 @@
...
 setup(
     name="pypiserver",
     description="A minimal PyPI server for use with pip/easy_install.",
-    long_description=open("README.rst").read(),
+    long_description=read_file("README.rst"),
     version=get_version(),
     packages=["pypiserver"],
     package_data={"pypiserver": ["welcome.html"]},
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
+    python_requires=">=3.6",
```
mplanchard, line 39: Good catch on this! Forgot to update the metadata when I dropped the last few versions.

mplanchard (approved):
This is awesome, thanks!

Issue #343, opened 2020-10-09
docker run with default config fails on v1.4.1


alex-m-humai:
Here is the command:

```
$ docker run pypiserver/pypiserver:v1.4.1
Unable to find image 'pypiserver/pypiserver:v1.4.1' locally
v1.4.1: Pulling from pypiserver/pypiserver
df20fa9351a1: Already exists 
36b3adc4ff6f: Already exists 
7eaf559fa1af: Already exists 
ff42851d39f6: Already exists 
8ba142e34a43: Already exists 
7b8d554bdc47: Pull complete 
d68d634278f4: Pull complete 
9aded7aa4123: Pull complete 
c547a38b5626: Pull complete 
Digest: sha256:abd275109b3bfa451d48275b7777e6eee17a0012a60d224b3bffdf9225ef00d3
Status: Downloaded newer image for pypiserver/pypiserver:v1.4.1
Using custom CMD: 
sh: : unknown operand
Error: while trying to list root(/home/pypiserver/packages): [Errno 2] No such file or directory: '/home/pypiserver/packages'

```

this works with v1.3.2

elfjes:
There indeed appears to be an issue with the new docker entrypoint, which sets a default package directory when no arguments are given. Currently the check results in an error. I've created a PR to address this issue

mplanchard:
Thanks for the report, and thanks @elfjes for the fix!

mplanchard:
The fix has been released in v1.4.2. Please let us know if you run into any more issues.

PR #344, opened 2020-10-10
Fix docker entrypoint script, improve docker build speed/caching

elfjes:
The current docker entrypoint.sh has a test `[ "$@" = "" ]`, which is not a valid test and actually results in an error when $@ is empty, which is the cause of #343

There are a few possible fixes
* `[ -z "$@" ]`
* `[ $# -eq 0 ]`

however, since all we want to do is point pypi-server to a default directory that is currently not the home directory, it is easier to just assign that directory as the home directory. This PR fixes that

furthermore, this PR addresses the following:
* clean up the docker context by creating a whitelist instead of a blacklist in `.dockerignore`
* speed up the docker build by copying only what we need in the build stage (currently, by copying the the full docker context into the build stage, cache invalidation may be unnecessarily triggered, which results in longer build times for consecutive builds)
* The entrypoint has a `set -o pipefail`. While this works with busbox sh, it doesn't work with bourne shell. This may lead to strange behaviour when someone tries to run the entrypoint outside a busybox/docker environment. I propose we just switch to bash inside the docker container. This also enhances the experience of a user exec'ing into a running container.


Dockerfile
```diff
@@ -54,8 +55,11 @@ COPY entrypoint.sh /entrypoint.sh
 # Use a consistent user and group ID so that linux users
 # can create a corresponding system user and set permissions
 # if desired.
-RUN addgroup -S -g 9898 pypiserver \
-    && adduser -S -u 9898 -G pypiserver pypiserver \
+RUN apk add bash \
+    && rm -rf /var/cache/apk/* \
+    && rm -rf /tmp/* \
+    && addgroup -S -g 9898 pypiserver \
+    && adduser -S -u 9898 -G pypiserver pypiserver --home /data\
```
mplanchard, line 62: This is clever. I like this solution.

mplanchard (approved):
Actually pulled it down and tested it this time. Looks good!

Issue #325, opened 2020-07-30
On modernization, code quality and contributions.


Luttik:
Let me start by saying that it is amazing that after almost 10 years this library is still the most popular (if you look at the github stars at least) way of hosting your own pypiserver. 

That said it does scare me that at this point @mplanchard is the only active maintainer to this project. Especially since the only comparable system that is actively maintained, devpi, is in a similar position with @fschulze being virtually the only contributor in the last 2 years (although the code frequency is much higher there.

## I'd like to see an effort to make pypiserver:
1. Easier to contribute to.
2. Easier to extend.
3. Making cloud hosting a breeze (e.g. relying less on local file storage, since persistent storage can be an issue). I understand that this seems less relevant than 1 & 2 but I believe this is vital if we want pypiserver to stay relevant.

## I think this requires a few changes / additions:
1. A clear overview of the PEPs and other requirements that pypiserver must adhere to. (this is essential to see what changes can and cannot be made)
2. Dropping support for old Python versions (say anything before 3.7) 
3. Adding code-quality tools (e.g. black)
4. Refactoring strange patterns (like having most of the code in `__init__.py` files and having bottle copied rather than imported. 
5. (more optional but would be very nice) replace bottle with a modern alternative say flask or fastapi.
6. Gradually introduce typehints (this will make diving into the code so much easier).

Personally I'd be willing to help achieve these goals and make pypiserver ready for the future.

aadith-pm:
I'd certainly like to help out if there is going to be a push for any of these. We've been setting up pypiserver and it's helped a lot so far. Surely want to contribute and give back.

aadithpm:
I'd certainly like to help out if there is going to be a push for any of these. We've been setting up pypiserver and it's helped a lot so far. Surely want to contribute and give back.

> Refactoring strange patterns (like having most of the code in init.py files and having bottle copied rather than imported.

Personally think this is a great starting point because this doesn't seem like good practise in my limited experience, but I can see why this was done initially.

dee-me-tree-or-love:
That's nice, all hands up for it. However, let's keep in mind the feasibility and urgency of this in the light of increased possibilities of package hostings provided as a service (e.g. GitLab made their PyPi server also available for users https://gitlab.com/help/user/packages/pypi_repository/index.md).

If you could transform your changes into todo items:   
- [ ] like this: `- [ ] Dropping support for...`  
Would be very handy to link it up to the next pull requests that would fix them up.

[...]

Issue #253, opened 2019-05-10
Alternative Storage Backends


reallyquickturtle:
Allow to store packages in cloud storage services such as AWS S3. 

This can be particularly useful when deploying on cloud infrastructure that has no persistent filesystem available (ECS, Heroku, etc.)

tomsaunders-code:
@mplanchard or @ankostis is this something you would be interested in supporting? I think may be able to help contribute to here.

The NPM community has a really nice private package server called Verdaccio that features this kind of plugin based storage backend.

https://github.com/verdaccio/verdaccio
https://github.com/Remitly/verdaccio-s3-storage

jayvdb:
I would be interested in a local backend that re-uses the real pip cache. c.f. https://github.com/jerith/landinggear/pull/2

deemetree-at-satelligence:
If anyone would be willing to collaborate and kick it off I would be happy to join up as well, looking in particular for options to conveniently use it with Google Cloud Storage in App Engine. (Possibly I have misread or not found a solution that already exists)

mplanchard:
This is definitely something we'd like to support! It's one of the major things I'd like to get into pypiserver this year.

deemetree-at-satelligence:
@mplanchard I have created a simple solution for the time being as a plugin to manage files between App Engine and Cloud Storage and would then submit it as a pull request if you are interested! Would much appreciate comments and collaboration on that further.

mplanchard:
@deemetree-at-satelligence yep, I'd love to work on incorporating that into pypiserver and am happy to collaborate w/you on that! If more discussion is needed before opening a PR, feel free to open a new issue to discuss it, or go ahead and open a PR if you'd like 🎉

deemetree-at-satelligence:
@mplanchard sorry, was a bit swamped this month with other tasks, but am picking it up! I will open up a WIP pull request soon

mplanchard:
Hey, no rush at all. I'm in the same boat 99.9% of the time 😄

Luttik:
Any news on this? I would love this functionality.

mplanchard:
[...]

PR #320, opened 2020-07-08
Add support for `data-requires-python`

Hyask:
This patch improves support for PEP 503 by adding the optional attribute `data-requires-python` when the info is available.
See #296 for more details.

No tests provided yet, but we can discuss adding some, even if I don't how you would like them implemented as it now requires more than just a filename.

PR #348, opened 2020-10-15
Refactor storage operations into separate Backend classes

elfjes:
Following the discussion in #253 and #325 I've created a first iteration on what a `Backend` interface could look like and how the current file storage operations may be refactored into this interface. It goes from the following principles

* `app.py` talks only to `core.py` with regards to package operations
* at configuration time, a `Backend` implementation is chosen and created for the lifetime of the configured app
* `core.py` proxies requests for packages to this `Backend()`
* The `Backend` interface/api is defined through three things
  * methods that an implementation must implement
  * methods that an implementation may override if it knows better than the defaults
  * the `PkgFIle` class that is (should be) the main carrier of data
* where possible, implementation details must be hidden from concrete `Backend`s to promote extensibility

Other things I've done in this PR:
* I've tried to talk about packages and projects, rather than files and prefixes, since these are the domain terms PEP503 uses, and imho it's also more clear what it means
* Better testability of the `CacheManager` (no more race conditions when `watchdog` is installed during testing)
* Cleanup some more Python 2 code
* Started moving away from  `os.path` and `py.path` in favour of `pathlib`

Furthermore I've created a `plugin.py` with a sample of how I think plugin system could look like. This sampIe assumes we use `argparse`  and allows for the extension of cli arguments that a plugin may need. I think the actual implementation of such a plugin system is beyond the scope of this PR, but I've used it as a target for the Backend refactoring. If requested, I'll remove it from this PR.

The following things still need to be done / discussed. These can be part of this PR or moved into their own, separate PRs
- [ ] Simplify the `PgkFile` class. It currently consists of a number of attributes that don't necessarily belong with it, and not all attributes are aptly named (imho). I would like to minimalize the scope of `PkgFile` so that its only concern is being a data carrier between the app and the backends, and make its use more clear.
- [ ] Add a `PkgFile.metadata` that backend implementations may use to store custom data for packages. For example the current `PkgFile.root` attribute is an implementation detail of the filestorage backends, and other Backend implementations should not be bothered by it.
- [ ] Use `pathlib` wherever possible. This may also result in less attributes for `PkgFile`, since some things may be just contained in a single `Path` object, instead of multtiple strings.
- [ ] Improve testing of the `CacheManager`.


mplanchard:
Awesome! Just FYI I have what will probably be a long day of work ahead and
some plans for this evening, so I probably won’t get to review this until
Friday evening, but I’m looking forward to it!

elfjes:
Sure no worries. Looking forward to your feedback :smile:

mplanchard:
Got a chance to give this a quick overview this evening, and I really like the direction! I have an in-progress branch to replace all usage of the old config with the stuff from #339 with all tests passing. You can see that [here](https://github.com/pypiserver/pypiserver/compare/config-replace). It seems like that could mesh really well with the approach taken here. I think we would update the `RunConfig` to expose a `backend` attribute, rather than having it expose an `iter_packages` method.

I've got no major concerns with what I've seen so far, but I'll be sure to make time to give this a more thorough review over the course of this week.

In regards to the things you mentioned as potentially needing more work:

> Simplify the PgkFile class. It currently consists of a number of attributes that don't necessarily belong with it, and not all attributes are aptly named (imho). I would like to minimalize the scope of PkgFile so that its only concern is being a data carrier between the app and the backends, and make its use more clear.

I agree with this goal! I don't feel strongly either way on whether it should be a part of this PR, but I definitely think it's worth exploring

> Add a PkgFile.metadata that backend implementations may use to store custom data for packages. For example the current PkgFile.root attribute is an implementation detail of the filestorage backends, and other Backend implementations should not be bothered by it.

Yeah, and having some access to metadata will also give us a route towards e.g. fully supporting PEP 503 and the `data_requires` attribute (a la #320).

> Use pathlib wherever possible. This may also result in less attributes for PkgFile, since some things may be just contained in a single Path object, instead of multtiple strings.

I've been working in this direction in my branch as well, so definitely no issues with seeing more of it.

> Improve testing of the CacheManager.

Yeah the existing watchdog cache had really minimal tests, so anything here is an improvement

---

Regarding getting both of our streams of work going harmoniously, we can either get this merged, and then I can rebase and update on top of this, or vice-versa. I'll probably go ahead and open a PR with that branch tonight or tomorrow, just so that we can better look at how they might work together, but there's no rush to get that one in, and I'm happy to do the work of rebasing if we get this one in first.

pypiserver/core.py
```diff
@@ -97,279 +87,87 @@ def auth_by_htpasswd_file(htPsswdFile, username, password):
...
+    return uri
 
-def find_packages(pkgs, prefix=""):
-    prefix = normalize_pkgname(prefix)
-    for x in pkgs:
-        if prefix and x.pkgname_norm != prefix:
-            continue
-        yield x
 
+def with_digester(func: t.Callable[..., t.Iterable[PkgFile]]):
```
mplanchard, line 129: to keep the decorator from overriding the underlying type, you can use a generic like:

```py
DigesterFunc = t.TypeVar("DigesterFunc", bound=t.Callable[..., t.Iterable[PkgFile]])

def with_digester(func: DigesterFunc) -> DigesterFunc:
  ...
```

you may need a `# type: ignore` where you're returning your wrapper function inside the decorator, but by telling mypy that you're returning a type of the same type that was put into the function, you allow it to infer that the resulting decorated function takes the same arguments as the original

pypiserver/manage.py
```diff
@@ -171,11 +169,11 @@ def update(pkgset, destdir=None, dry_run=False, stable_only=True):
 def update_all_packages(
     roots, destdir=None, dry_run=False, stable_only=True, blacklist_file=None
 ):
-    all_packages = itertools.chain(*[core.listdir(r) for r in roots])
+    all_packages = itertools.chain(*[listdir(r) for r in roots])
```
mplanchard, line 172: I know you didn't write the original, but since we're touching it, we can keep this (potentially) lazy by updating like:

```suggestion
    all_packages = itertools.chain.from_iterable(listdir(r) for r in roots)
```

tests/test_app.py
```diff
@@ -111,6 +105,15 @@ def welcome_file_all_vars(request, root):
...
 
+def add_file_to_root(root, filename, content=""):
+    root.join(filename).write(content)
+    if isinstance(core.backend, CachingFileBackend):
+        cache = core.backend.cache_manager.invalidate_root_cache(root)
+
+
+# @pytest.mark.xfail(
+#     ENABLE_CACHING, reason="race condition when caching is enabled"
+# )
```
mplanchard, line 116: dead code?

pypiserver/_app.py
```diff
@@ -104,7 +91,7 @@ def root():
     fp = request.custom_fullpath
 
     try:
-        numpkgs = len(list(packages()))
+        numpkgs = len(list(core.get_all_packages()))
```
mplanchard, line 94: It would be nice if we added a "package_count" or similar method to the backend so that, for example, we wouldn't actually have to parse all the files into `PkgFile` instances in order to get their count, and could instead just nab all files whose names match a regex or something. It would also be useful with imagined remote backends, where pulling all the files might be a problem

pypiserver/_app.py
```diff
@@ -292,59 +276,62 @@ def simpleindex():
...
+    packages = sorted(
+        core.get_all_packages(),
         key=lambda x: (os.path.dirname(x.relfn), x.pkgname, x.parsed_version),
     )
+
     links = [
-        (f.relfn_unix, urljoin(fp, f.fname_and_hash(config.hash_algo)))
-        for f in files
+        (pkg.relfn_unix, urljoin(fp, pkg.fname_and_hash)) for pkg in packages
     ]
```
mplanchard, line 333: Could we try replacing this list comprehension (`[]`) with a generator comprehension (`()`)? We definitely need to allocate a list for sorting, but here I think we may be able to be lazy and just pass a generator into the template call, which should save some RAM on systems handling a lot of packages

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+            self.digest = self.digester(self)
+        hashpart = f"#{self.digest}" if self.digest else ""
+        return self.relfn_unix + hashpart
+
+
+class Backend:
+    def __init__(self, config: Configuration):
+        self.hash_algo = config.hash_algo
+
+    def get_all_packages(self) -> t.Iterable[PkgFile]:
```
mplanchard, line 71: Could we explicitly say that this returns `t.Iterator[PkgFile]` instead of `t.Iterable`? I would like to guide implementations into being lazy with returning all the packages wherever possible

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        """Implement this method to return an Iterable of all packages (as
+        PkgFile objects) that are available in the Backend.
+        """
+        raise NotImplementedError
+
+    def add_package(self, filename: str, fh: t.BinaryIO):
```
mplanchard, line 77: bikeshed0: I think `stream` or `bytestream` may be a more generic name for this parameter and will make more sense for implementations not using the FS

we may also explicitly want to note the return type being `None` here (and for other methods with no return value)
```suggestion
    def add_package(self, filename: str, fh: t.BinaryIO) -> None:
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        the file's content. To convert the package into an actual file on disk,
+        run `as_file(filename, fh)`.
+        """
+        raise NotImplementedError
+
+    def remove_package(self, pkg: PkgFile):
+        """Remove a package from the Backend"""
+        raise NotImplementedError
+
+    def exists(self, filename) -> bool:
```
mplanchard, line 90:
```suggestion
    def exists(self, filename: str) -> bool:
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        """Does a package by the given name exist?"""
+        raise NotImplementedError
+
+    def digest(self, pkg: PkgFile):
+        if self.hash_algo is None:
+            return None
+        return digest_file(pkg.fn, self.hash_algo)
+
+    def get_projects(self) -> t.Iterable[str]:
```
mplanchard, line 99: same note here re: iterator
```suggestion
    def get_projects(self) -> t.Iterator[str]:
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        """Return an iterable of all (unique) projects available in the store
+        in their PEP503 normalized form. When implementing a Backend class,
+        either use this method as is, or override it with a more performant
+        version.
+        """
+        normalized_pkgnames = set()
+        for x in self.get_all_packages():
+            if x.pkgname:
+                normalized_pkgnames.add(x.pkgname_norm)
```
mplanchard, line 108: Is this check for the `pkgname` necessary? It seems like, since it' a required positional argument to the `PkgFile`, we should be able to rely on it (and the normalized name) always being present, in which case we could do something like

```py
yield from set(pkg.pkgname_norm for pkg in self.get_all_packages())
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        return normalized_pkgnames
+
+    def find_project_packages(self, project: str) -> t.Iterable[PkgFile]:
```
mplanchard, line 111:
```suggestion
    def find_project_packages(self, project: str) -> t.Iterator[PkgFile]:
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        Backend class, either use this method as is, or override it with a
+        more performant version.
+        """
+        return (
+            x
+            for x in self.get_all_packages()
+            if normalize_pkgname(project) == x.pkgname_norm
+        )
+
+    def find_version(self, name, version) -> t.Iterable[PkgFile]:
```
mplanchard, line 123:
```suggestion
    def find_version(self, name, version) -> t.Iterator[PkgFile]:
```

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @property
+    def fname_and_hash(self):
+        if self.digest is None:
+            self.digest = self.digester(self)
+        hashpart = f"#{self.digest}" if self.digest else ""
+        return self.relfn_unix + hashpart
+
+
+class Backend:
```
mplanchard, line 67: It would probably be helpful to implementors to include an example or two of what we expect each of these methods' inputs to be in their docstrings

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+            self.cache_manager.listdir(r, listdir) for r in self.roots
+        )
+
+    def digest(self, pkg: PkgFile):
+        return self.cache_manager.digest_file(
+            pkg.fn, self.hash_algo, digest_file
+        )
+
+
+def as_file(fh: t.BinaryIO, destination: PathLike):
```
mplanchard, line 173: bikeshed: `as_file()` or `to_file()` feels to me like the function will be returning something file-like. What about `write_file()`?

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        return itertools.chain.from_iterable(listdir(r) for r in self.roots)
+
+    def add_package(self, filename: str, fh: t.BinaryIO):
+        as_file(fh, self.roots[0].joinpath(filename))
+
+    def remove_package(self, pkg: PkgFile):
+        os.remove(pkg.fn)
+
+    def exists(self, filename):
+        # TODO: Also look in subdirectories?
```
mplanchard, line 150: I think currently we're recursing into root directories, so we may want to do so here too

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        return any(root.joinpath(filename).exists() for root in self.roots)
+
+
+class CachingFileBackend(SimpleFileBackend):
+    def __init__(
+        self, config: Configuration, roots: t.List[PathLike], cache_manager
```
mplanchard, line 156: We may want to define either a base class or a [protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) for the cache_manager

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+        either use this method as is, or override it with a more performant
+        version.
+        """
+        return filter(
+            lambda pkg: pkg.pkgname == name and pkg.version == version,
+            self.get_all_packages(),
+        )
+
+
+class SimpleFileBackend(Backend):
```
mplanchard, line 135: I think mypy will probably want us to define types for the subclasses. Since this is a new file, we can add it to the typechecks in `tox.ini` to make sure it catches anything we miss

pypiserver/backend.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    Reads and digests a file according to specified hashing-algorith.
+
+    :param hash_algo: any algo contained in :mod:`hashlib`
+    :return: <hash_algo>=<hex_digest>
+
+    From http://stackoverflow.com/a/21565932/548792
+    """
+    blocksize = 2 ** 16
+    digester = getattr(hashlib, hash_algo)()
```
mplanchard, line 220: let's pull in my update from #349, which essentially just replace this line with
```suggestion
    digester = hashlib.new(hash_algo)
```
to support a wider variety of hashing algorithms

mplanchard (commented):
Found some time this evening for a first pass!

pypiserver/core.py
elfjes: good idea! :)

pypiserver/manage.py
elfjes: Sure, I wanted to touch manage.py as little as possible, but this totally makes sense

tests/test_app.py
elfjes: good catch, will remove

pypiserver/_app.py
elfjes: a `package_count` method sounds good, but I'm not sure about adding "query" or filter functionality. This would require additional complexity from each backend (including said imagined ones) without adding apparent value. What would be a use case for having such filtering functionality in a `package_count` method?

pypiserver/_app.py
elfjes: Yeah sure :)

Good catch on the `fname_and_hash`. Apparently no test caught that

pypiserver/backend.py
elfjes: I appreciate the sentiment, but I'm not quite sure about restricting the interface to `Iterator`. `Backend.get_projects()` for example returns a `set`. It doesn't look right to me to have to return an `iter(set(...))`. Guiding people to have as much lazy iterables as possibl can also be achieved through other means, such as documentation.

pypiserver/backend.py
elfjes: Yeah this extra check is quite redundant

pypiserver/backend.py
elfjes: Yeah, good idea. Documentation has to be improved for the whole "implement your own backend"-flow in general :)

pypiserver/backend.py
elfjes: Alas, no `Protocol`s in python<3.8. I see that I put the type_hint for cache_manager in the `__init__` assignment rather that the function signature, so I'm changing that. 

Did you mean to add a sort of CacheManager interface/ base implementation that different backends may consume and/or extend? This might be an idea, but I'm afraid that there is not going to be much common requirements between the different backends, since most of the logic will lie in cache invalidation, which is going to be very backend-specific

pypiserver/backend.py
elfjes: I'm not too familiar with mypy and I'm not quite sure what you mean with this. A subclass is a type, right? Or do you mean something else?
I'll add the new file(s) to the `tox.ini` and see what needs be done :)

elfjes:
Thanks for the review, much appreciated :)

> Got a chance to give this a quick overview this evening, and I really like the direction! I have an in-progress branch to replace all usage of the old config with the stuff from #339 with all tests passing. You can see that [here](https://github.com/pypiserver/pypiserver/compare/config-replace). It seems like that could mesh really well with the approach taken here. I think we would update the `RunConfig` to expose a `backend` attribute, rather than having it expose an `iter_packages` method.

Yeah I was also thinking of having a `--backend` attribute which picks a backend. See also `plugin.py` in which I philosophized a bit on how we can support selection and configuration of different backends

> > Simplify the PgkFile class. It currently consists of a number of attributes that don't necessarily belong with it, and not all attributes are aptly named (imho). I would like to minimalize the scope of PkgFile so that its only concern is being a data carrier between the app and the backends, and make its use more clear.
> 
> I agree with this goal! I don't feel strongly either way on whether it should be a part of this PR, but I definitely think it's worth exploring
Maybe it's best to give it it's own PR, so that we don't have to make this one larger than it already is :)

> > Add a PkgFile.metadata that backend implementations may use to store custom data for packages. For example the current PkgFile.root attribute is an implementation detail of the filestorage backends, and other Backend implementations should not be bothered by it.
> 
> Yeah, and having some access to metadata will also give us a route towards e.g. fully supporting PEP 503 and the `data_requires` attribute (a la #320).
Yeah I realized that there are two types of metadata that we must distinguish. One is the PEP503 metadata that we need for full support of this PEP. The other is geared towards extra data that a specific backend implementation may need to keep track of it's packages, such as a package `root` for a FileBackend, a bucket name for an imagined MultiBucketGCSBackend or a packages primary key in a database, and have a designated place to store such additional attributes. My thought initially was with the latter, but we need the former as well. 

> Regarding getting both of our streams of work going harmoniously, we can either get this merged, and then I can rebase and update on top of this, or vice-versa. I'll probably go ahead and open a PR with that branch tonight or tomorrow, just so that we can better look at how they might work together, but there's no rush to get that one in, and I'm happy to do the work of rebasing if we get this one in first.

Yeah sure, I'll finish up this PR and then create a new one for the additional work to ease the integration pain a little bit

pypiserver/_app.py
mplanchard: Sorry, I just meant a count without any custom filtering, but I expressed that poorly. Where I was talking about a regex, I essentially just meant "count every file that looks like a package based on the filename"

pypiserver/backend.py
mplanchard: Yeah, that seems reasonable. `Iterator` and `Iterable` can generally be consumed the same way. I just struggle to see any case where we'd want to be eager rather than lazy with getting packages. Even pulling from the local filesystem, it's probably better to be lazy to avoid being RAM hogs. I'm good with either approach though

pypiserver/backend.py
mplanchard: Yep, I just meant that we would need to type these methods for mypy to be happy, and that mypy will enforce that they match the base class types to satisfy the liskov substitution principle

pypiserver/backend.py
mplanchard: > Alas, no Protocols in python<3.8

This is true, but I think that if we want to use it, you can get it for earlier pythons by installing [typing-extensions](https://pypi.org/project/typing-extensions/).

> Did you mean to add a sort of CacheManager interface/ base implementation that different backends may consume and/or extend

Yep, that was the idea, so that you could potentially have like a `WatchdogCachingFileBackend` and a `WhateverOtherCachingFileBackend`, so long as the mangers share the same interface. It seems like all
we need for that at the moment is `listdir()` and `digest_file()`, so it might not be too bad to specify an abstract interface

elfjes:
I've added a `--backend` option to the cli that by default automatically chooses between the caching and the simple file backend.

It took me a while to figure out how to exactly add the argument and make it work with the `Config.default_with_overrides` function that is used in the tests. Which got me thinking, if we want to add pluggable backends, these new backends may want to add new command line arguments. How would we deal with them in the currently design of the (Run)Config, which has fixed supported arguments? I guess we'd have to go back to some form of dynamic config objects, although that's probably a discussion for a different issue/PR

Speaking of pluggable backends, @mplanchard when you have time, could you give me your feedback on the proposal I made for plugins in `plugin.py`?

Also, if you're ok with the current status of this PR, we can merge it, so that I can start working on the other things in a separate PR.

mplanchard:
> It took me a while to figure out how to exactly add the argument and make it work with the Config.default_with_overrides function that is used in the tests

There is definitely still room for improvement in making these obvious/easy to add to. 

> Which got me thinking, if we want to add pluggable backends, these new backends may want to add new command line arguments. How would we deal with them in the currently design of the (Run)Config, which has fixed supported arguments? I guess we'd have to go back to some form of dynamic config objects, although that's probably a discussion for a different issue/PR

I think the way we would approach this would be to have a dynamic portion of the config object, e.g. `backend_config`, which is a generic namespace. Plugins would need to provide some function to parse arguments and return a namespace, which we'd use to populate that property. Or something like that :shrug:

elfjes:
yeah that could work. Or like a `**custom` argument to the config object that gets converted to a namespace. Or just kept as a dictionary

pypiserver/cache.py
```diff
@@ -4,10 +4,24 @@
...
+try:
+    from watchdog.observers import Observer
+
+    ENABLE_CACHING = True
+
+except ImportError:
+
+    Observer = None
+
+    ENABLE_CACHING = False
```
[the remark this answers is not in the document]
mplanchard: Ah I see you added this already!

pypiserver/config.py
```diff
@@ -719,6 +736,22 @@ def auther(uname: str, pw: str) -> bool:
...
+            "cached-dir": CachingFileBackend,
+        }
+        # if arg not in available_backends:
+        #     raise argparse.ArgumentTypeError(
+        #         f"Value must be one of {', '.join(available_backends.keys())}"
+        #     )
+
+        backend = available_backends[arg]
+
+        return BackendProxy(backend(self))
```
mplanchard, line 753: Should we put this in the `_ConfigCommon` so that it's available for the update command too?

pypiserver/config.py
```diff
@@ -504,9 +520,12 @@ def get_parser() -> argparse.ArgumentParser:
 
 
 TConf = t.TypeVar("TConf", bound="_ConfigCommon")
+BackendFactory = t.Callable[["Configuration"], Backend]
 
 
 class _ConfigCommon:
+    hash_algo: t.Optional[str] = None
```
mplanchard, line 527: Should we move the `hash_algo` init argument to this common config?

pypiserver/core.py
```diff
@@ -1,154 +1,68 @@
...
+        "version",  # The package version as a string
+        "fn",  # The full file path
+        "root",  # An optional root directory of the file
+        "relfn",  # The file path relative to the root
+        "replaces",  # The previous version of the package (used by manage.py)
+        "pkgname_norm",  # The PEP503 normalized project name
+        "digest",  # The file digest in the form of <algo>=<hash>
+        "relfn_unix",  # The relative file path in unix notation
+        "parsed_version",  # The package version as a tuple of parts
+        "digester",  # a function that calculates the digest for the package
```
mplanchard, line 38: thanks for this documentation!

pypiserver/core.py
```diff
@@ -1,154 +1,68 @@
...
         self.pkgname_norm = normalize_pkgname(pkgname)
         self.version = version
-        self.parsed_version = parse_version(version)
+        self.parsed_version: tuple = parse_version(version)
         self.fn = fn
         self.root = root
         self.relfn = relfn
         self.relfn_unix = None if relfn is None else relfn.replace("\\", "/")
         self.replaces = replaces
+        self.digest = None
```
mplanchard, line 63: Should this be `digester` to match the `__slots__`?

pypiserver/plugin.py
```diff
@@ -0,0 +1,41 @@
...
+    and return a Backend instance. It may be the class constructor or a factory
+    function to construct a Backend object
+
+In the future, the plugin callable may be called with additional keyword
+arguments, so a plugin should accept a **kwargs variadic keyword argument.
+"""
+from pypiserver.backend import SimpleFileBackend, CachingFileBackend
+from pypiserver import get_file_backend
+
+DEFAULT_PACKAGE_DIRECTORIES = ["~/packages"]
```
mplanchard, line 24: Can we pull this from `config.DEFAULTS`?

pypiserver/config.py
elfjes: Yes, that makes sense to me. I think the main reason I didn't do that yet, was because it currently requires `hash_algo` to be part of the same config and that one's on the RunConfig. But if you're fine with moving that to common, then I see no reason to not do that.

pypiserver/core.py
elfjes: Haha, I mostly put it there for my own sanity until I had the chance to refactor `PkgInfo`

pypiserver/core.py
elfjes: Both digest and digester are part of slots. I do see that I need to assign `self.digester=None` in `__init__` to be consistent

pypiserver/plugin.py
elfjes: So in my opinion, if we're doing this as plugins, then the package root is a config setting specific to the file backends. Which would also mean that the cli argument should not be part of the global config, but provided by the plugin. Which is why I added that definition there, and not take it from the global config

mplanchard: That does make sense. I just want to avoid needing to update this twice if we update it (although the likelihood that we'll ever change this is pretty low tbf)

pypiserver/config.py
mplanchard: Yep, now that we need it in both the run and update configs, we might as well move it

elfjes:
Hi @mplanchard,

It's been a while, how are you? What do you think about the status of this PR? As far as I'm concerned it's ready to be merged. I've processed most, if not all of your feedback so far. Are there still any blocking things?

pypiserver/config.py
```diff
@@ -565,13 +597,21 @@ def log_level(self) -> int:
...
+
+        available_backends: t.Dict[str, BackendFactory] = {
+            "auto": get_file_backend,
+            "simple-dir": SimpleFileBackend,
+            "cached-dir": CachingFileBackend,
+        }
+        # if arg not in available_backends:
+        #     raise argparse.ArgumentTypeError(
+        #         f"Value must be one of {', '.join(available_backends.keys())}"
+        #     )
```
mplanchard, line 610: Dead code?

mplanchard (approved):
Sorry for being slow on this. Q4 and the beginning of Q1 have been crazy at my day job. Let's get it in!

mplanchard:
Leaving it to you to merge. Let me know if you don't have permissions, and I'll hit the button instead

dee-me-tree-or-love:
@mplanchard @elfjes if no rights to merge, I can do also! 
(sorry for my low participation lately as well... also quite overloaded with various tasks on the job. @mplanchard, what do you also think, actually is it okay for me to also be merging/approving when I get a moment for it, maybe I will be poking for your call upfront in the future still?)

mplanchard:
> @mplanchard, what do you also think, actually is it okay for me to also be merging/approving when I get a moment for it, maybe I will be poking for your call upfront in the future still

Yeah I think so! For bigger stuff, we should probably both review, but for small things I'm okay with either of us approving and merging. And feel free to tag me for a review if there's anything at all you're uncertain about or would like me to weigh in on. If there's anything time sensitive, bug me in Zulip. It's harder for me to ignore real time chat than GH issues 🤪.

pypiserver/config.py
elfjes: Ah thanks, yeah I thought I might need to check that here, but it's actually picked up by argparse

elfjes:
I removed that dead code you mentioned, so I guess it needs reapproval. I don't have a merge button so if any of you @mplanchard @dee-me-tree-or-love can merge it, I'd be much obliged :)

mplanchard (approved):

PR #350, opened 2020-10-26
Add PypiserverApp app factory

mplanchard:
This adds a `PypiserverApp` app factory to make the process of
instantiating the app, providing a config, etc. more straightforward
and easy to reason about. The factory is instantiated with a config
and a Bottle instance, and the `_initialize_app()` method adds routes
and hooks. Methods on the class have access to the config as needed.

This avoids the need to mutate the top-level global scope of the `_app`
module, along with the need to pop it out of `sys.modules`. While
the end result is the same, hopefully this is a bit easier for newcomers
to reason about.

Since we no longer need to do odd things with the import path, the
`_app` module has been renamed to `pypiserver_app`.


elfjes:
Hey, looks interesting :)

I am wondering a bit though, how big is the use case of requesting more than one app? Is it only for testing, or are there other use cases? If it's only for testing we can just reset the globals for every test case, right? Pypiserver is an application, not a library, and in our application runtime we only ever instantiate an app only once.

Lastly, and I assume that this is not going to be a problem, is this thing compatible with indirect usage, such as through gunicorn (#352)? I don't have enough gunicorn experience to judge that.

nice touch to add a .pylintrc. From the amount of exceptions, I guess that there is quite a long way to go :)

mplanchard:
I see you crossed out the question about use-cases, but I wanted to answer it b/c it's a good question. I have no way of knowing how common it is, but I could see the use-case for running different pypiserver apps via the API interface in the same process for the sake of exposing different package directories with different permissions on different ports. For example, someone could expose low-sensitivity packages and allow uploads from wherever in the company in one directory, and have another directory where downloads only are allowed. This same use-case could be served (probably more effectively too) with different processes, but it's something I could see people doing.

> Lastly, and I assume that this is not going to be a problem, is this thing compatible with indirect usage, such as through gunicorn (#352)?

Yep, I am fairly certain that this will behave in exactly the same way as the current implementation.

pypiserver/pypiserver_app.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def __init__(self, config: RunConfig, app: Bottle) -> None:
+        """Construct a new Pypiserver application."""
+        self.app = app
+        self.config = config
+        self.auth = functools.partial(auth, config)
+        self._initialize_app()
+        # Add a reference to our config onto the bottle instance,
+        # so anyone using the app() constructor can get it if needed
+        self.app._pypiserver_config = config
```
dee-me-tree-or-love, line 45: Hey, this looks great all of it, but if I may propose, how about making the factory and the wrapper class separate? Curious for arguments for and against. Then that would take the `_initialize_app()` method outside of this class. 

I think it could make it easier like this for example: 

```python
class PypiserverApp:
    def __init__(self, config: RunConfig) -> None:
        # without any reference to Bottle here
        self.config = config
        self.auth = functools.partial(auth, config)
   
     def log_request(self) -> None:
        log.info(self.config.log_req_frmt, request.environ)
     
     # and so forth with the other hook methods...

class PypiserverAppFactory:

    @classmethod
    def build_app(cls, config: RunConfig, app: Bottle) -> Bottle:
        pypiserver_app = PypiserverApp(config)
        return cls._initialize_app(app, pypiserver_app)

   @staticmethod
    def _initialize_app(app: Bottle, pypiserver_app) -> Bottle:
        app.hook("before_request")(pypiserver_app.log_request)
        app.hook("before_request")(pypiserver_app.parse_request_path)
        # and so on...
        app._pypiserver_config = config
        return app
```
Then potentially, later this would allow extending the `pypiserver` with ohther backends for people who could want it, if instead of `app: Bottle` the parameter will be `builder: AppBuilder` or so that will be an instance of a strategy to initialize an app of a particular backend given the pypiserver wrapper. But that's a bit too far off maybe. Let me know :)

mplanchard: Yeah, I like this idea. I'll take a look at updating the PR sometime soon!

Issue #352, opened 2020-10-27
Pip install packages with the docker image


ArcLightSlavik:
I have a docker-compose of pypi which runs with the default server:

```
  pypi:
    image: &pypi_version pypi:20.10.1
    build:
      context: services/pypi
      dockerfile: Dockerfile
    environment:
      PORT: "8000"
    ports:
      - "8000:8000"
    volumes:
      - ~/code/monitor/services/pypi:/app
    command: -P . -p 8000 -a . /data/packages
```

I want to user gunicorn server so I changed my `command` to:
```
command: -P . -p 8000 -a . --server gunicorn /data/packages
```
But I get a `ModuleNotFoundError: No module named 'gunicorn'` when running it.

I've tried running pip install similar to the pypiserver Dockerfile. It does return `Successfully installed gunicorn-20.0.4` but the module error persists.

```
FROM pypiserver/pypiserver

RUN apk add --no-cache --virtual .build-deps \
        build-base \
        libffi-dev \
    && mkdir /install \
    && python -m pip install --no-warn-script-location \
                --prefix=/install \ 
                gunicorn

EXPOSE 8000
```

ArcLightSlavik:
Got a similar result by doing
```
FROM python:3.7.7-stretch

ADD requirements.txt /app/
RUN pip install -r /app/requirements.txt

VOLUME /data/packages
EXPOSE 8000

CMD ["gunicorn", "-b", ":8000", "pypiserver:app(root='/data/packages', port=8000, password_file='.', authenticated='.')"]
```

elfjes:
So, there's something funky going on. This should work (but it doesn't, I will explain the reasons later):
```Dockerfile
FROM pypiserver/pypiserver
RUN pip install gunicorn
```
```bash
docker build -t pypiserver:gunicorn . # or however you'd want to name your docker image
docker run -p 8000:8000 pypiserver:gunicorn -P . -p 8000 -a . --server gunicorn
```
But like I said, it doesn't work. It fails with a message `pypi-server: error: unrecognized arguments: -P -a . --server gunicorn /data/packages`. Apparently, in its current form, the pypi-server cli / bottle is incompatible with gunicorn. When bottle calls the gunicorn server, gunicorn runs `argparse` and parses the command line arguments. The command line arguments are obviously incompatible with gunicorn, as they were meant for pypi-server. At this moment I don't know exactly how to fix this.

[...]

PR #353, opened 2020-11-03
Add --diff flag to black formatter check

dee-me-tree-or-love:
# What

This makes the output more explicit. 
Instead of just a check like this: 

```bash
$ black --check .
would reformat /home/travis/build/pypiserver/...
```

Will do a check like this showing the possible changes.

```bash
$ black --diff --check .
....
     expected_results = [
         mock_removal_change_event_class(difference=set()),
-        mock_addition_change_event_class(difference={'c'}),
+        mock_addition_change_event_class(difference={"c"}),
     ]
     mock_last_local_snapshot.return_value = expected_snapshot
     mock_local_contents.return_value = expected_local_contents
     result = list(storage_client.get_change_events())
     assert expected_results == result
would reformat tests/...
```

## Remarks:

- Check this for details: https://github.com/psf/black#command-line-options
- Not sure if that's required, but I thought it could be handy. Feel free to close :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,4 @@ jobs:
   include:
     - python: 3.8
       install: pip install -U black
-      script: black --check .
+      script: black --diff --check .


mplanchard (approved):
Yeah, this looks useful! I think you'll need to pull in current master and resolve conflicts from #352, and then this is good to go!

mplanchard:
I went ahead and pulled this into #361!

PR #354, opened 2020-11-07
Refactor test_server to increase speed

elfjes:
I gave `test_server.py` some much needed attention. This file now take ~30 seconds on my machine to run (down from 130 seconds), and I cleaned up the code ~~a little~~. Let's see how this goes in CI


tests/test_server.py
```diff
@@ -11,169 +11,194 @@
...
-    global _port
-    _port += 1
-    return _port
-
-
-Srv = namedtuple("Srv", ("proc", "port", "package"))
-
-
-def _run_server(packdir, port, authed, other_cli=""):
+@contextlib.contextmanager
```
mplanchard, line 40: nice use of a contextmanager for this

tests/test_server.py
```diff
@@ -11,169 +11,194 @@
...
 # ######################################################################
 # Fixtures & Helper Functions
 # ######################################################################
 
 
-_BUFF_SIZE = 2 ** 16
-_port = 8090
-SLEEP_AFTER_SRV = 3  # sec
+CURRENT_PATH = Path(__file__).parent
+ports = itertools.count(10000)
```
mplanchard, line 36: Could we make this something like `range(8000, 10000)`, so that it won't try to bind privileged ports?

tests/test_server.py
```diff
@@ -11,169 +11,194 @@
...
+def _authed_server(server_root):
+    with run_server(server_root, authed=True) as srv:
+        yield srv
 
-protected_port = 8082
+
+@pytest.fixture
+def authed_server(_authed_server):
+    yield _authed_server
+    clear_directory(_authed_server.root)
```
mplanchard, line 179: Nice use of the module and function-scoped fixtures to keep a long-lived server while ensuring a fresh state for each test.

tests/test_server.py
```diff
@@ -187,281 +212,155 @@ def _run_pip(cmd):
...
     try:
-        yield
+        yield pypirc_path
     finally:
         if old_pypirc:
-            pypirc_path.write(old_pypirc)
+            pypirc_path.write_text(old_pypirc)
         else:
-            pypirc_path.remove()
+            pypirc_path.unlink()
```
mplanchard, line 243: At some point we should see if we can figure out a workaround for needing to (even if temporarily) overwrite the pypirc file of whoever's running tests. Not critical at the moment, since it's been that way for ages, but it would be nice

tests/test_server.py
```diff
@@ -187,281 +212,155 @@ def _run_pip(cmd):
...
 
 # ######################################################################
 # Tests
 # ######################################################################
 
+all_servers = [
+    ("open_server", "open_pypirc"),
+    ("authed_server", "authed_pypirc"),
+    ("partial_authed_server", "authed_pypirc"),
+]
```
mplanchard, line 281: Nice :+1:

mplanchard (approved):
Huge improvement, thank you!

tests/test_server.py
elfjes: Yeah I looked at that. With twine we can specify a specific location for the `.pypirc` file, but for `setup.py` we can't. I can think of a few options though:
- we could change the $HOME when running `setup.py` to point to a different location
- uploading packages with `setup.py` is deprecated so we may consider just removing those tests

tests/test_server.py
elfjes: So currently it starts at 10000, so it shouldn't run into any privileged ports, although a range object also works I guess :)

mplanchard: oh I see! that's my mistake then. I have never used `count`, and I assumed it worked like range

PR #356, opened 2020-11-08
Fix version incompatibility that breaks twine in docker test

### LINKED ISSUE #355: Incompatible dependencies in CI
The docker tests have just broken in CI under python<3.8. This is due to an upgrade of `keyring` to `21.5.0` (which is a `twine` dependency), that requires `importlib-metadata>=1;python_version<3.8` (`keyring<21.5.0` has a `importlib-metadata; python_version<3.8` requirement). The travis-ci container already has `importlib-metadata==0.18`, which is compatible with the earlier version of keyring, but not with the latest version. Twine (ie. `pkg_resources`) now gives an error during the docker tests (see: https://travis-ci.org/github/pypiserver/pypiserver/jobs/742144229).

I would propose a twofold solution to this:
- use the new pip resolver to find and fix the incompatibiltiy
- run the docker tests only on a higher python version. This would then be a separate stage in CI, which also would decouple this test from the tox tests, and the test is only being run once, instead of multiple time (one for every tox environment). Right now the docker test is quite fast, so it may or may not be faster than running it once in a separate stage (due to setup time of a new ci stage), but if we improve the docker tests (#346), the tests may be increase in time

elfjes:
fixes #355 

- use the new pip resolver for dependeny installation
- run the docker tests in a separate stage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ python:
   - pypy3
 
 install:
-  - pip install --use-feature 2020-resolver -U setuptools twine pip sphinx tox tox-travis
+  - pip install -U pip
+  - pip install --use-feature=2020-resolver -U setuptools twine sphinx tox tox-travis
 
 script:
   - tox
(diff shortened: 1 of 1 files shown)


mplanchard (approved):
:+1:

Issue #359, opened 2020-11-15
Improve docker image


elfjes:
There are currently a number of issues with the docker image
 - Requires root permissions (which leads to #357)
 - Using the docker image for anything else than docker run (this currently also throws an error in the docker tests (when the script calls `docker run pypiserver:test --help` , but this error is ignored)
- It uses the default wsgiref server which is non-theaded and not production ready. We should pick another.
- I noticed that the pypiserver uses /data as a root, but /data/packages as a volume, which means that new package uploads may not be persisted
- Currently `PORT` is used as an environment variable to customize the port. While environment variables for configuration of containers is considered good practice, I believe we should make this variable more specifc, ie `PYPISERVER_PORT`. We can then also expand this to other configuration settings, and maybe not limit it to only the docker image.

Together with the lack of testing (#346) I think the docker image release of pypiserver should be given some more attention. I think the `entrypoint.sh` currently does more harm than good and we should try and remove it again (while keeping docker-compose working), or improve on it drastically

mplanchard:
> Requires root permissions (which leads to #357)

I don't think this is true. It just needs permissions to change ownership on the `/data/packages` volume, right?

> Using the docker image for anything else than docker run (this currently also throws an error in the docker tests (when the script calls docker run pypiserver:test --help , but this error is ignored)

Yep this is definitely a problem.

> It uses the default wsgiref server which is non-theaded and not production ready. We should pick another.

In general people should be using the docker image like the do the application, which would mean putting it behind a more appropriate nginx or similar ingress for production workloads.

> I noticed that the pypiserver uses /data as a root, but /data/packages as a volume, which means that new package uploads may not be persisted

`/data` is used as the WORKDIR and the `$HOME` for the pypiserver user, and `/data/packages` is declared as a VOLUME. I'm pretty sure that anything declared as a VOLUME will be persisted unless the container is run with `--rm` or unless the volumes are explicitly removed.

> Currently PORT is used as an environment variable to customize the port. While environment variables for configuration of containers is considered good practice, I believe we should make this variable more specifc, ie PYPISERVER_PORT. We can then also expand this to other configuration settings, and maybe not limit it to only the docker image.

This seems like a good idea.

> I think the entrypoint.sh currently does more harm than good and we should try and remove it again (while keeping docker-compose working), or improve on it drastically

I agree with this. As I mentioned in the other thread, #330 has been I think more more pain than gain.

elfjes:
[...]

PR #365, opened 2021-02-02
Docker improvements

mplanchard:
This addresses much of what was brought up in #359. Specifically, it:

- Significantly improves testing for the Docker image, adding a
  `docker/test_docker.py` file using the regular pytest machinery to
  set up and run docker images for testing
- Hopefully addresses a variety of permissions issues, by being explicit
  about what access pypiserver needs and asking for it, only erroring
  if that access is not available
  - Requires RX permissions on `/data` (R to read files, X to list files
    and to be able to cd into the directory. This is important since
    `/data` is the `WORKDIR`)
  - Requires RWX permissions on `/data/packages`, so that we can list
    packages, write packages, and read packages.
  - When running in the default configuration (as root on Linux or
    as the pypiserver-named rootish user on Mac), with no volumes
    mounted, these requirements are all satisfied
  - Volume mounts still must be readable by the pypiserver user (UID
    9898) in order for the container to run. However, we now error early
    if this is not the case, and direct users to a useful issue.
  - If the container is run as a non-root, non-pypiserver user (e.g.
    because someone ran `docker run --user=<user_id>`, we try to run
    pypiserver as that user). Provided that user has access to the
    necessary directories, it should run fine.
- Fixes issues with running help and similar commands
- Updates the Docker image to use `PYPISERVER_PORT` for port
  specification, while still falling back to `PORT` for backwards
  compatibility
- Moves some docker-related things into a `/docker` directory
- Adds a `Makefile` for building a test fixture package sdist and wheel,
  so that test code can call `make mypkg` and not need to worry about it
  potentially building multiple times

The only issue #359 raises that's not addressed here is the one of
running pypiserver in the Docker container using some non-default server
for performance. I would like to do some benchmarking before deciding on
what to do there.


mplanchard:
Tagging @elfjes for review

docker/test_docker.py
```diff
@@ -0,0 +1,589 @@
...
+    """Wait for the container to be available."""
+    for _ in range(60):
+        try:
+            httpx.get(f"http://localhost:{port}").raise_for_status()
+        except (httpx.RequestError, httpx.HTTPStatusError):
+            time.sleep(1)
+        else:
+            return
+
+    # If we reach here, we've tried 20 times without success.
```
elfjes, line 91: 60 times?

docker/test_docker.py
```diff
@@ -0,0 +1,589 @@
...
+    if any(
+        ln.strip().startswith("pypiserver-mypkg")
+        for ln in out.decode().splitlines()
+    ):
+        proc = subprocess.Popen(("pip", "uninstall", "-y", "pypiserver-mypkg"))
+        proc.communicate()
+
+
+@pytest.fixture(scope="session")
+def session_cleanup() -> t.Iterator[None]:
```
elfjes, line 111: It looks like this fixture is never used

docker/test_docker.py
```diff
@@ -0,0 +1,589 @@
...
+
+        Returns the container ID.
+        """
+        proc = subprocess.Popen(
+            (
+                "docker",
+                "run",
+                "--rm",
+                "--publish",
+                "8080:8080",
```
elfjes, line 280: Can we not bind on 8080 on the host but on a random port, someone running the tests may have something else running on that port already

docker/test_docker.py
```diff
@@ -0,0 +1,589 @@
...
+    def test_runs_as_pypiserver_user(self, image: str) -> None:
+        """Test that the default run uses the pypiserver user."""
+        proc = subprocess.Popen(
+            (
+                "docker",
+                "run",
+                "--rm",
+                "--detach",
+                "--publish",
+                "8080:8080",
```
elfjes, line 229: Can we not bind on 8080 on the host but on a random port, someone running the tests may have something else running on that port already

docker/test_docker.py
```diff
@@ -0,0 +1,589 @@
...
+        """View the welcome page."""
+        resp = httpx.get("http://localhost:8080")
+        assert resp.status_code == 200
+        assert "pypiserver" in resp.text
+
+
+class TestAuthed:
+    """Test basic pypiserver functionality in a simple unauthed container."""
+
+    HOST_PORT = 8081
```
elfjes, line 389: also, random port please :)

fixtures/mypkg/setup.py
```diff
@@ -0,0 +1,10 @@
+"""A simple setup file for this test package."""
+
+from setuptools import setup
+
+setup(
+    name="pypiserver_mypkg",
+    description="Test pkg",
+    version="1.0.0",
+    packages=["mypkg"],
+)
```
elfjes, line 10: what about just using the `centodeps` approach from the unittests to create a simple package?

elfjes (commented):
Nice! :tada:  Much better than before. Lots of subprocess plumbing though, but I guess that can't be helped :man_shrugging:

docker/test_docker.py
mplanchard: whoops, didn't update the comment when I updated the number. good cach!

docker/test_docker.py
mplanchard: Ah I meant to make it autouse and forgot to. Another good catch

fixtures/mypkg/setup.py
mplanchard: Personally I like having it checked in, because it more clearly mirrors the "real world," and would also allow us to eventually add fixture packages that do things like include C extensions or utilize alternate build systems (like poetry) more easily. It's largely six of one and half dozen of the other, but I think it's nice when experimenting to be able to go look at the file, run `setup.py` on it directly, upload it to a locally running pypiserver, etc.

mplanchard:
> Lots of subprocess plumbing though, but I guess that can't be helped

yeah I usually make a subprocess wrapper, and didn't here because I was like, "hey, it's probably overkill," but then the file wound up being like 99% subprocess calls, so I went ahead and added my usual `run()` method

PR #369, opened 2021-02-02
📝 Updated the setup.py file

Pratham-Dave:
Updated the setup file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setup(
     name="pypiserver",
     description="A minimal PyPI server for use with pip/easy_install.",
     long_description=read_file("README.rst"),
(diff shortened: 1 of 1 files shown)


mplanchard (commented):
What are the reasons for the version changes in this file?

In addition to setting the long description content type, we'll still need to convert the README file to markdown

dee-me-tree-or-love:
Hey @Pratham-Dave thanks for the proposal! I share the thoughts of @mplanchard regarding the changes, are you still around to give some more motivation about this? :) It seems to me that it slightly runs ahead of the train and we'd need to make a few internal changes before this can be safely merged. I think we will be slowly moving towards the modernisation changes in the repository, so your input would also be very helpful.

Issue #286, opened 2019-12-12
Server randomly hangs


allanmachado1:
I'm running pypiserver using the Docker image version 1.3.1 with minimal change. My current Dockerfile is the following:

    FROM pypiserver/pypiserver:v1.3.1
    ADD htpasswd /htpasswd
    CMD ["--disable-fallback", "-P", "/htpasswd", "-a", "update,download,list", "packages"]

Sometimes the server just hangs. 
It can happen when I'm installing something from the index using pip or pushing a new package to it using twine. In both cases the process just hangs indefinitely. Usually it works when I try to run it again immediately after, but once in a while I'm forced to restart the container running pypiserver.

I've already tried image version 1.2.7 and gunicorn as the server, hoping this behavior would change, but it just keeps happening.

I don't currently have many packages in it (about 20).

mplanchard:
Thanks for the report! Are you running `pypiserver` directly or are you reverse proxying with a proper webserver like nginx or similar?

mplanchard:
The current Docker image (`latest` or `v1.3.2`) includes some significant changes, including running in Python 3 rather than Python 2. Could you check to see if this is still happening?

ric79:
Hello, I just noticed that the server hangs after browsing http://XXXX:8082/packages/

Here there is my docker run

```
sudo docker run -d -p 8082:8080 \
       --name pypiserver \
       --volume /root/pypiserver/auth/.htpasswd:/data/.htpasswd \
       --volume /root/pypiserver/packages:/data/packages \
       --user $(id -u $(whoami)) \
       pypiserver/pypiserver:v1.3.2 \
       -vvvvvv --disable-fallback --overwrite --passwords .htpasswd -a update,download,list packages
```
Logs are ok: 

```
2020-01-15 18:44:45,636|pypiserver.bottle|INFO|139829829008712|Bottle v0.13-dev server starting up (using AutoServer())...
2020-01-15 18:44:45,636|pypiserver.bottle|INFO|139829829008712|Listening on http://0.0.0.0:8080/
2020-01-15 18:44:45,636|pypiserver.bottle|INFO|139829829008712|Hit Ctrl-C to quit.
```

and `python3 setup.py sdist upload -r local` works. 
However after browsing packages it hangs:
```
$ python3 setup.py sdist upload -r local
...
removing 'mysimplessh-0.0.1' (and everything under it)
running upload
Submitting dist/mysimplessh-0.0.1.tar.gz to XXX:8082/
[...]

PR #371, opened 2021-02-07
Gunicorn/gevent docker, log fixes, cache busting

### LINKED ISSUE #358: Logs not showing when running from docker
I am in ubuntu 20.04.
When i start pypiserver with command
```
docker run --rm -p 8080:8080 -v /tmp/packages:/data/packages pypiserver/pypiserver -P . -a .
```
and then, try to install some packages
```
pip install --no-cache-dir --force-reinstall --exists-action=w example-pkg --trusted-host 127.0.0.1 --index-url http://127.0.0.1:8080/simple/
```
It doesn't show any logs from console. But if i stopped the container, it will show all accumulated logs at once.
image

Notice in the images above, before blank lines, I download one package, But it doesn't show any logs. After i stop container, it show 3 logs at once.

mplanchard:
Updates the Docker configuration to use the gunicorn server with gevent
workers by default. Hopefully this will resolve the symptoms seen when
running pypiserver with the non-production grade wsgiref server as
reported in #286.

Making this happen brought a few other issues to light, which are also
addressed here.

- Docker log output not immediately being flushed to stdout (#358):
  resolved by setting the `PYTHONUNBUFFERED` env var to `t` in the
  docker container
- When the WSGIRef server is selected, its access logs are written
  directly to stderr, rather than going through the logging machinery:
  resolved by adding a new `WsgiHandler` class and passing in to
  bottle's `run()` method when running the wsgi server. This required a
  new `ServerCheck` class to determine whether the wsgi server is
  selected when the `auto` option is used
- When using `gunicorn` along with the watchdog cache, package uplaods
  were not being picked up by the watcher. Updated the `add_package`
  and `remove_package` methods on the `CachingFileBackend` to bust the
  cache
- The Dockerfile was not really amenable to local development, since any
  change to code led to all the dependencies being reinstalled. Updated the
  file so that dependencies can be cached as a separate image from code.

---

Resolves #358

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -117,7 +117,9 @@ fi
 
 
 if [[ "$*" == "" ]]; then
-    CMD=("run" "-p" "${PYPISERVER_PORT:-$PORT}")
+    # Use the gunicorn server by default, since it's more performant than
+    # bottle's default server
+    CMD=("run" "-p" "${PYPISERVER_PORT:-$PORT}" "--server" "gunicorn")
 else
     # this reassigns the array to the CMD variable
     CMD=( "${@}" )
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -6,21 +6,27 @@ from __future__ import print_function
 import logging
 import sys
 import typing as t
+from pathlib import Path
+from wsgiref.simple_server import WSGIRequestHandler
 
 import functools as ft
 from pypiserver.config import Config, UpdateConfig
+
+if t.TYPE_CHECKING:
+    import pypiserver.bottle  # noqa pylint: disable=unused-import
 
 
 log = logging.getLogger("pypiserver.main")
 
 
 def init_logging(
-    level=logging.NOTSET,
-    frmt=None,
-    filename=None,
+    level: int = logging.NOTSET,
+    frmt: str = None,
+    filename: t.Union[str, Path] = None,
     stream: t.Optional[t.IO] = sys.stderr,
-    logger=None,
-):
+    logger: logging.Logger = None,
+) -> None:
+    """Configure the specified logger, or the root logger otherwise."""
     logger = logger or logging.getLogger()
     logger.setLevel(level)
 
@@ -36,14 +42,132 @@ def init_logging(
         logger.addHandler(handler)
 
 
-def main(argv=None):
+class WsgiHandler(WSGIRequestHandler):
+    """A simple request handler to configure logging."""
+
+    # The default `FixedHandler` that bottle's `WSGIRefServer` uses does not
+    # log in a particularly predictable or configurable way. We'll pass this
+    # in to use instead.
+    def address_string(self) -> str:  # Prevent reverse DNS lookups please.
+        # This method copied directly from bottle's `FixedHandler` and
+        # maintained on the Chesterton's fence principle (i.e. I don't know
+        # why it's important, so I'm not going to get rid of it)
+        return self.client_address[0]
+
+    def log_message(
+        self, format: str, *args: t.Any  # pylint: disable=redefined-builtin
+    ) -> None:
+        """Log a message."""
+        # The log_message method on the `HttpRequestHandler` base class just
+        # writes directly to stderr. We'll use its same formatting, but pass
+        # it through the logger instead.
+        log.info(
+            "%s - - [%s] %s\n",
+            self.address_string(),
+            self.log_date_time_string(),
+            format % args,
+        )
+
+
+class ServerCheck:
+    """Methods to check for a server's presence in the import path."""
+
+    # Because we need to potentially override behavior for different servers
+    # by passing options into bottle's `run` method, and because those options
+    # aren't necessarily compatible between servers, we need to be able to guess
+    # which server bottle will pick if the user specifies the `auto` server
+    # type. These methods' names match the `__name__` property of the server
+    # adapters in bottle, so we can filter the adapters on the `AutoServer`
+    # adapter to figure out which ones are present. Since bottle uses the
+    # first server to successfully import from that same list, this should
+    # be sufficient to give us good certainty as to which server it will run.
+
+    # pylint: disable=missing-function-docstring
+    # pylint: disable=invalid-name
+    # pylint: disable=import-outside-toplevel
+    # pylint: disable=unused-import
+
+    @staticmethod
+    def WaitressServer() -> bool:
+        try:
+            import waitress  # noqa: F401
+
+            return True
+        except ImportError:
+            return False
+
+    @staticmethod
+    def PasteServer() -> bool:
+        try:
+            import paste  # noqa: F401
+
+            return True
+        except ImportError:
+            return False
+
+    @staticmethod
+    def TwistedServer() -> bool:
+        try:
+            import twisted.web  # noqa: F401
+
+            return True
+        except ImportError:
+            return False
+
+    @staticmethod
+    def CherryPyServer() -> bool:
+        try:
+            import cheroot.wsgi  # noqa: F401
+
+            return True
+        except ImportError:
+            try:
+                import cherrypy.wsgiserver  # noqa: F401
+
+                return True
+            except ImportError:
+                return False
+
+    @staticmethod
+    def WSGIRefServer() -> bool:
+        try:
+            import wsgiref  # noqa: F401
+
+            return True
+        except ImportError:
+            return False
+
+    # pylint: enable=missing-function-docstring
+    # pylint: enable=invalid-name
+    # pylint: enable=import-outside-toplevel
+    # pylint: enable=unused-import
+
+
+def guess_auto_server() -> "t.Type[pypiserver.bottle.ServerAdapter]":
+    """Guess which server bottle will use for the auto setting."""
+    # pylint: disable=import-outside-toplevel
+    import pypiserver.bottle  # pylint: disable=redefined-outer-name
+
+    # Return the first ServerAdapter in `AutoServer.adapters` whose
+    # corresponding method in `ServerCheck` returns True. If none of the
+    # servers import correctly, `next()` will raise a `StopIteration` error.
+    return next(
+        filter(
+            lambda s: getattr(ServerCheck, s.__name__)(),
+            pypiserver.bottle.AutoServer.adapters,
+        )
+    )
+
+
+def main(argv: t.Sequence[str] = None) -> None:
     """Application entrypoint for pypiserver.
 
     This function drives the application (as opposed to the library)
     implementation of pypiserver. Usage from the commandline will result in
     this function being called.
     """
-    import pypiserver
+    # pylint: disable=import-outside-toplevel
+    import pypiserver  # pylint: disable=redefined-outer-name
 
     if argv is None:
         # The first item in sys.argv is the name of the python file being
@@ -84,25 +208,52 @@ def main(argv=None):
     from pypiserver import bottle
 
     bottle.debug(config.verbosity > 1)
-    bottle._stderr = ft.partial(
+    bottle._stderr = ft.partial(  # pylint: disable=protected-access
         _logwrite, logging.getLogger(bottle.__name__), logging.INFO
     )
 
     # Here `app` is a Bottle instance, which we pass to bottle.run() to run
     # the server
     app = pypiserver.app_from_config(config)
+
+    if config.server_method == "gunicorn":
+        # When bottle runs gunicorn, gunicorn tries to pull its arguments from
+        # sys.argv. Because pypiserver's arguments don't match gunicorn's,
+        # this leads to errors.
+        sys.argv = ["gunicorn"]
+
+    wsgi_kwargs = {"handler_class": WsgiHandler}
+
+    if config.server_method == "auto":
+        expected_server = guess_auto_server()
+        extra_kwargs = (
+            wsgi_kwargs if expected_server is bottle.WSGIRefServer else {}
+        )
+        log.debug(
+            "Server 'auto' selected. Expecting bottle to run '%s'. "
+            "Passing extra keyword args: %s",
+            expected_server.__name__,
+            extra_kwargs,
+        )
+    else:
+        extra_kwargs = wsgi_kwargs if config.server_method == "wsgiref" else {}
+        log.debug(
+            "Running bottle with selected server '%s'", config.server_method
+        )
+
     bottle.run(
         app=app,
         host=config.host,
         port=config.port,
         server=config.server_method,
+        **extra_kwargs,
     )
 
 
 def _logwrite(logger, level, msg):
     if msg:
         line_endings = ["\r\n", "\n\r", "\n"]
-        for le in line_endings:
+        for le in line_endings:  # pylint: disable=invalid-name
             if msg.endswith(le):
                 msg = msg[: -len(le)]
         if msg:
--- a/pypiserver/__main__.py
+++ b/pypiserver/__main__.py
@@ -1,8 +1,7 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 """Entrypoint for pypiserver."""
 
-from __future__ import print_function
-
+import importlib
 import logging
 import sys
 import typing as t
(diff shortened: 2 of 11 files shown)


pypiserver/__main__.py
```diff
@@ -36,14 +42,132 @@ def init_logging(
...
+    # first server to successfully import from that same list, this should
+    # be sufficient to give us good certainty as to which server it will run.
+
+    # pylint: disable=missing-function-docstring
+    # pylint: disable=invalid-name
+    # pylint: disable=import-outside-toplevel
+    # pylint: disable=unused-import
+
+    @staticmethod
+    def WaitressServer() -> bool:
```
elfjes, line 91: wanna use a helper like?
```python
import importlib
def can_import(module):
    try:
        importlib.import_module(module)
        return True
    except ImportError:
        return False
```

pypiserver/__main__.py
```diff
@@ -36,14 +42,132 @@ def init_logging(
...
+
+    # Return the first ServerAdapter in `AutoServer.adapters` whose
+    # corresponding method in `ServerCheck` returns True. If none of the
+    # servers import correctly, `next()` will raise a `StopIteration` error.
+    return next(
+        filter(
+            lambda s: getattr(ServerCheck, s.__name__)(),
+            pypiserver.bottle.AutoServer.adapters,
+        )
+    )
```
elfjes, line 159: We're assuming a number of available servers here (the implemented methods of `ServerCheck` so we're not really decoupled from whatever is defined in `AutoServer.adapters`. Since we also use the result only for identification of a `server_method` (see my other comment below), we may just hard code a couple of `server_method` strings in here, which could simplify this function to something like this:

```python
def guess_auto_server():
    server_methods = [('waitress', 'waitress'), ('paste', 'paste'), ('twisted', 'twisted.web'), ('cherrypy', 'cherrypy.wsgiserver'), ('wsgiref','wsgiref')]
    return next(s[0] for s in server_methods if can_import(s[1]))
```

Also, do you think it's a good idea to slap a try/except StopIteration on it to give a more useful error message? I guess the chances of it failing with a StopIteration are slim because the `wsgiref` package should always be available, but idk, lettng StopIteration propagate seems kinda strange.

docker/entrypoint.sh
```diff
@@ -117,7 +117,9 @@ fi
 
 
 if [[ "$*" == "" ]]; then
-    CMD=("run" "-p" "${PYPISERVER_PORT:-$PORT}")
+    # Use the gunicorn server by default, since it's more performant than
+    # bottle's default server
+    CMD=("run" "-p" "${PYPISERVER_PORT:-$PORT}" "--server" "gunicorn")
```
elfjes, line 122: I know it's not technically part of this PR, but should we do a check here to see if `$PORT` is and give a deprecation warning if it is, before using it? (I guess we then also need to not set it in `Dockerfile`)

pypiserver/__main__.py
```diff
@@ -84,25 +208,52 @@ def main(argv=None):
...
 
     # Here `app` is a Bottle instance, which we pass to bottle.run() to run
     # the server
     app = pypiserver.app_from_config(config)
+
+    if config.server_method == "gunicorn":
+        # When bottle runs gunicorn, gunicorn tries to pull its arguments from
+        # sys.argv. Because pypiserver's arguments don't match gunicorn's,
+        # this leads to errors.
+        sys.argv = ["gunicorn"]
```
elfjes, line 223: How can we pass additional configuration to gunicorn? only through a `gunicorn.conf.py`?

And just out of curiosity: how is the `gunicorn.conf.py` passed as a config file in the docker container? Or does `gunicorn` pick it up automatically from the user's $HOME?

pypiserver/backend.py
```diff
@@ -167,6 +167,15 @@ def __init__(
 
         self.cache_manager = cache_manager or CacheManager()  # type: ignore
 
+    def add_package(self, filename: str, stream: t.BinaryIO) -> None:
+        write_file(stream, self.roots[0].joinpath(filename))
+        self.cache_manager.invalidate_root_cache(self.roots[0])
+
+    def remove_package(self, pkg: PkgFile) -> None:
+        if pkg.fn is not None:
```
elfjes, line 175: is this ever called with `None`?

pypiserver/backend.py
```diff
@@ -167,6 +167,15 @@ def __init__(
...
+            os.remove(pkg.fn)
```
elfjes, line 176: I think we need to catch `FileNotFoundError` here to be safe. (Or maybe even `OSError`, although that might be too broad a catch)

elfjes (commented):
nice :) I like the idea of getting `gunicorn` to work through the normal `--server` arg. However, I am a bit concerned about patching `sys.argv`. It seems like its a big hammer that may produce unwanted side effects. Although, I don't have any other idea on how to do it. And we're running an app, not a library. So it's probably fine. :sweat_smile: 

do you think it's possible to get a test to verify that we're actually running gunicorn?

furthermore, do we need some tests to check new behaviour of `main()` with regards to the server/extra_kwargs selection?

pypiserver/__main__.py
```diff
@@ -84,25 +208,52 @@ def main(argv=None):
...
+            "Passing extra keyword args: %s",
+            expected_server.__name__,
+            extra_kwargs,
+        )
+    else:
+        extra_kwargs = wsgi_kwargs if config.server_method == "wsgiref" else {}
+        log.debug(
+            "Running bottle with selected server '%s'", config.server_method
+        )
+
```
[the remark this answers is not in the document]
mplanchard: We could do that, but it would involve an additional layer of translation, since at the bottle layer, the `.adapters` list is the source of truth for servers might be chosen, and it's already an array of `ServerAdapter` instances. So, if we wanted to use a string, we'd have to map those adapters back to our strings, which seems like it'd be adding an extra potential source of errors. Ideally bottle would provide some method to _get_ the auto server instance itself, but the `AutoServer` only provides a `run()` method unfortunately.

pypiserver/__main__.py
mplanchard: Yeah, this seems like a good idea. I thought about this, but leaned against it at the time because I wanted to be sure we were importing in the same way that bottle is, with direct statements. My understanding is that `importlib.import_module` will do the same thing though, so this should work too!

pypiserver/__main__.py
mplanchard: The intent is to remain coupled to what's defined in `.adapters`, since that's bottle's source of truth on what can run. I'd like to be sure that if we pull in a bottle update that updates that list, we wind up with an error because we don't have a method for it in `ServerCheck`. 

We can catch the error and raise something more specific, for sure. My thinking was in line with yours, like it _should_ be impossible because `wsgiref` is in the standard library, but we might as well be safe.

docker/entrypoint.sh
mplanchard: I struggled a bit with this, because what if they have the env var `$PORT` set for some other reason or as part of something else they've got running on a derived docker container?

pypiserver/__main__.py
mplanchard: Yeah, this is another thing I struggled with. There are a couple of ways as it stands that folks could configure gunicorn. One is setting the `GUNICORN_CMD_ARGS` env var, another is by mounting a new gunicorn config file, and another would be by setting certain env vars directly. But, I couldn't figure out any sensible way to have them passed via the commandline.

The `gunicorn.conf.py` is picked up automatically when it's in the current working directory when you're running gunicorn. We might consider adding something like a `--gunicorn-conf` option to allow people to specify some config file rather than using gunicorn's default search mechanisms, but since they can already do this with `GUNICORN_CMD_ARGS="-c my_config_file.py"` it seemed like it wasn't strictly necessary

pypiserver/backend.py
mplanchard: Honestly I copied these straight out of the `SimpleFileBackend` and added the cache manager calls without thinking too much about it. I probably should update to calling `super().add_package(...)` and then update the ones there to address these concerns

mplanchard:
> nice :) I like the idea of getting gunicorn to work through the normal --server arg. However, I am a bit concerned about patching sys.argv. It seems like its a big hammer that may produce unwanted side effects. Although, I don't have any other idea on how to do it. And we're running an app, not a library. So it's probably fine. 😅

I don't love it either. IMO the gunicorn programmatic interface shouldn't be trying to parse argv at all! If we were doing anything that running `bottle.run()` here, I'd consider saving the original value of `argv` and patching it back, but since we need to update it for the `run()` call, and we exit immediately after the `run()` call, it didn't seem worth it

> do you think it's possible to get a test to verify that we're actually running gunicorn?

Yeah, I think this shouldn't be too hard. I definitely verified myself manually (we get different logs), but it's worth a quick test

> furthermore, do we need some tests to check new behaviour of main() with regards to the server/extra_kwargs selection?

yep, can do

mplanchard:
Also just noting re: the `argv` thing that if someone is using pypiserver itself as a lib, they should be using one of the `app` constructors (`app` or `app_from_config`) in `__init__.py` and then passing that WSGI application to some server themselves, so they won't hit the code that patches `argv`

pypiserver/backend.py
mplanchard: I don't know if this ever gets called with `pkg.fn` being none. It's allowed to be none according to the type definition of `PkgFile`, but it does seem like we'd need something like this. I'll dig in a bit and see if this check is really necessary, or if we can maybe change the type of the `PkgFile` class

pypiserver/backend.py
mplanchard: So frustrating that python doesn't include which exceptions can be raised from a function in the stdlib documentation honestly

pypiserver/__main__.py
elfjes: Hmm, if I understand you correctly, you are worried that we might inadvertently run out of sync in the way we resolve `auto` and `bottle` does. The way this function works now, in CI it goes through all items in `Autoserver.adapters` because we don't install any third party servers in CI so it ends up at `WSGIRefServer` always.  When a new `ServerAdapter` is added or a name changes, we do not have any corresponding method in `ServerCheck` and we get an error in CI (currently StopIteration), so that we know something has changed. Whenever the tests are run on another machine, behaviour is undefined since any third party servers might be installed.

This kind of sounds like a very implicit way of testing whether something has changed in bottle. Might be better to write an explicit test for this, if this is what you're worried about. When using strings, this would be something like reverse-mapping `bottle.server_names` and checking our list of strings against whatever `AutoServer.adapters` maps to.

docker/entrypoint.sh
elfjes: Hmm, I see. What about not showing that warning when a $PYPISERVER_DISABLE_DEPRECATION_WARNINGS variable is set? or is that taking things too far :sweat_smile: ?

pypiserver/__main__.py
elfjes: Ah right, it's automatically picked up from the working dir, I had a look at the gunicorn docs, but couldn't find it there. Thanks :)

I'd say adding a gunicorn specific command line argument is not really the way to go. Maybe in the future there could be a way to pass arbitrary options to arbitrary servers, but I guess that's for another time :)

pypiserver/backend.py
elfjes: Haha, then it must've been my doing :roll_eyes: . I guess I did the None-check to make the typechecker happer.

I'm currently working on simplifying `PkgFile`, so will have a look at it as well

docker/entrypoint.sh
mplanchard: Hahaha, I think that's probably good actually, although then I'd have to go update the deprecation warning I'm showing for the old arg form as well

elfjes:
> > nice :) I like the idea of getting gunicorn to work through the normal --server arg. However, I am a bit concerned about patching sys.argv. It seems like its a big hammer that may produce unwanted side effects. Although, I don't have any other idea on how to do it. And we're running an app, not a library. So it's probably fine. sweat_smile
> 
> I don't love it either. IMO the gunicorn programmatic interface shouldn't be trying to parse argv at all! If we were doing anything that running `bottle.run()` here, I'd consider saving the original value of `argv` and patching it back, but since we need to update it for the `run()` call, and we exit immediately after the `run()` call, it didn't seem worth it

Right, yeah that limits the usefulness of restoring the patched values

> 
> > do you think it's possible to get a test to verify that we're actually running gunicorn?
> 
> Yeah, I think this shouldn't be too hard. I definitely verified myself manually (we get different logs), but it's worth a quick test
> 
> > furthermore, do we need some tests to check new behaviour of main() with regards to the server/extra_kwargs selection?
> 
> yep, can do

cool :+1:

pypiserver/__main__.py
mplanchard: I threw a few extra comments in around this in the most recent commits. It'll also be something we want to document as we work on getting the docs up to snuff for v2

pypiserver/__main__.py
mplanchard: Yep, you understand my concern perfectly. We vendor bottle, so we're insulated from unexpected changes, but I would at least like to know if something breaks. Mapping the servers to some other format seems (to me) like extra work for not a whole lot of gain, since we'd have to validate:

- our string mapping retains the same order as the bottle server mapping
- all of the bottle adapters are mapped to string representations
- we have no extra string representations that aren't represented in the bottle adapters

which I guess you could verify with something like `tuple(map(to_string_mapping, bottle.adapters)) == tuple(string_mapping_to_import_check_map)`.

We could do that in unit tests, which does reduce coupling to bottle at the application layer, and so is probably not a bad idea, but the coupling still exists anyway and remains intimately tied to bottle's adapter protocol. 

I'll play with it a bit and see what it feels like. If nothing else, I'll throw a TODO in to improve the decoupling in a future commit

pypiserver/backend.py
elfjes: Yeah I know, but I guess that'd be documentation that's very hard to maintain if it needs to list all possible exceptions.

also: I meant `IOError` earlier, not `OSError`. It's still probably to broad an exception group to (silently) handle

mplanchard:
Went ahead and also in the most recent commits included `waitress` as a dependency of the Docker container. I suspect that, regardless of how well we document things, folks who override args may not remember to set `--server gunicorn` as part of their override, and I think we probably want that to fall back to something other than wsgiref

pypiserver/backend.py
mplanchard: I added in a catch of `FileNotFoundError` with a log message and a pass, and a catch of `OSError`, which logs an exception and then re-raises :shrug:

docker/entrypoint.sh
elfjes: Yeah exactly. One would expect that to not only hide docker-specific deprecations, but also those coming from the python source

pypiserver/__main__.py
elfjes: something like this?

somewhere in `__main__.py`

```python
  server_methods = [('waitress', 'waitress'), ('paste', 'paste'), ('twisted', 'twisted.web'), ('cherrypy', 'cherrypy.wsgiserver'), ('wsgiref','wsgiref')]
```

somewhere in the tests:
```python
import pypiserver.bottle
def test_bottle_auto_resolution_order():
    adapter_rmap = {
        val: key for key, val in pypiserver.bottle.server_names.items()
    }
    assert [adapter_rmap[a] for a in pypiserver.bottle.AutoServer.adapters] == [
        s[0] for s in server_methods
    ]

mplanchard: Okay I added this (using an enum instead of strings) in a87e771. The tests wound up being a bit awkward because we need to do two separate import checks for cherry py, so it required deduping, but I think it's reasonable.

docker/entrypoint.sh
mplanchard: I would like to think a bit more about this and potentially do it in a separate PR, but I do think it's a good idea

pypiserver/__main__.py
elfjes: Hmm I don't quite understand the benefit of using Enum in this case. I get the idea, but the result doesn't look necessarily nice imho. But sure :)

I think you might be able to reduce the awkwardness by changing the signature of `_can_import` to `def _can_import(*names: str)` (add a for loop in body), let the generator expr be `(s for s, *i in AUTO_SERVER_IMPORTS if _can_import(*i))` and the IMPORT `(AutoServer.CherryPy, "cheroot.wsgi", "cherrypy.wsgiserver")` so you'd have unique entries in there

The assert statements seem like they can then be replaced by a simple `==` assertion on the AUTO_SERVER_IMPORTS first items lower str

mplanchard: For typing, I prefer having a defined enumeration of options than a general `str`, since the variants of a string are unlimited. Returning an enum allows callers to potentially verify statically that they've covered all the cases. I think the code itself is fine enough. The tests are a bit awkward, but nothing too crazy.

The asserts can't be replaced by a regular `==` because the enumeration names don't exactly match the `ServerAdapter` class names, by virtue of wanting to decouple them.

PR #379, opened 2021-02-19
Add option to allow only overwriting development releases

sparkyb:
I like the security of not allowing packages with the same version to be overwritten if I forget to bump the version number. However, I'd like to have an automated build job that publishes a development release any time I push to my development branch. Rather than have to remember to edit the version with every commit I push, I want this automated build to automatically add a .dev0 suffix to make sure it is publishing dev releases. I don't want it to have to determine the highest dev version number to increment and I don't mind overwriting these. I could just turn `--overwrite` on, but I thought that this was a reasonable use case to have an option that only allowed overwriting of dev releases. So I added a `--overwrite-dev` option that does that.

Internally, it writes to the same config option, so now the option can have 3 possible values: False (no overwrite), True (overwrite any package), "dev" (overwrite only dev releases). I'd be happy to switch it to a real enum if you'd prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,8 @@ optional arguments:
                         one of paste, cherrypy, twisted, or wsgiref.
   -o, --overwrite       Allow overwriting existing package files during
                         upload.
+  --overwrite-dev       Allow overwriting existing package files during upload,
+                        only for development releases.
   --welcome HTML_FILE   Use the contents of HTML_FILE as a custom welcome
                         message on the home page.
   --cache-control AGE   Add "Cache-Control: max-age=AGE" header to package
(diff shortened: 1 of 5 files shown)


mplanchard:
Thanks for this PR! Just FYI I'm extremely busy right now w/work and a few things that have come up outside of work, so it will be probably a couple of weeks before I can get back to reviewing PRs. I will get to them as soon as possible!

adouge:
Hey! This is another great feature I would love to see, as I'd been using it in a closed network without authentication, using overwrite protection 
But indeed, manually bumping .devN is a bit annoying...

Is there anything one could do to help move this further?


One issue comes to mind, how to specify "overwriteable" versions? 

Since Python (or semantic versioning) does not really differentiate between, say, alpha/beta/RC and "dev" versions

PR #380, opened 2021-02-24
Correct README example

qubidt:
The readme incorrectly specifies `watchdog` as the name of the extra to install. It's actually specified as `cache`. see: https://github.com/pypiserver/pypiserver/blob/df300de33d2ffe01f218396813433f23001dd76b/setup.py#L49


mplanchard (approved):
Great catch, thank you!

PR #395, opened 2021-07-19
added JSON Topic for use with micropython-upip

awachtler:
This patch adds a json-info url to pypiserver, which is needed for micropython-upip.

The change enables pypi-server to be used as local package server for micropython projects.

If server is queried like this: http://gretel-2:8080/pypiserver/json it returns this JSON data:

```
{
"info": {"version": "2.0.0.dev1"}, 
"releases": {
  "2.0.0.dev1": [{"url": "http://gretel-2:8080/packages/pypiserver-2.0.0.dev1.zip"}], 
  "2.0.0dev1aw": [{"url": "http://gretel-2:8080/packages/pypiserver-2.0.0dev1aw.zip"}]
 }
}
```

I would be happy, if this change can make it to the master branch.

Best Regards, Axel.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -8,6 +8,7 @@ import zipfile
 from collections import namedtuple
 from io import BytesIO
 from urllib.parse import urljoin, urlparse
+from json import dumps
 
 from pypiserver.config import RunConfig
 from . import __version__
@@ -361,6 +362,21 @@ def server_static(filename):
             return response
 
     return HTTPError(404, f"Not Found ({filename} does not exist)\n\n")
+
+@app.route("/:project/json")
+@auth("list")
+def json_info(project):
+    rv = {}
+    releases = []
+    for x in config.backend.find_project_packages(project):
+        r = {x.version: [{"url": "http://localhost:8080/packages/" + x.relfn_unix}]}
+        releases.append(r)
+    rv["releases"] = releases
+    # todo check if this really that simple
+    max_version = max([r for r in releases.keys()])
+    rv = {"info" : {"version": max_version}}
+    response.content_type = 'application/json'
+    return dumps(rv)
 
 
 @app.route("/:project")
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -366,15 +366,19 @@ def server_static(filename):
 @app.route("/:project/json")
 @auth("list")
 def json_info(project):
-    rv = {}
-    releases = []
-    for x in config.backend.find_project_packages(project):
-        r = {x.version: [{"url": "http://localhost:8080/packages/" + x.relfn_unix}]}
-        releases.append(r)
-    rv["releases"] = releases
-    # todo check if this really that simple
-    max_version = max([r for r in releases.keys()])
-    rv = {"info" : {"version": max_version}}
+    packages = sorted(
+        config.backend.find_project_packages(project),
+        key=lambda x: x.parsed_version, reverse=True
+    )
+    max_version = packages[0].version
+    releases = {}
+    req_url = request.url
+    for x in packages:
+        releases[x.version] = [{"url": urljoin(req_url, "../../packages/" + x.relfn)}]
+    rv = {
+        "info" : {"version": max_version},
+        "releases" : releases
+    }
     response.content_type = 'application/json'
     return dumps(rv)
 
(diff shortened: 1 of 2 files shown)


dee-me-tree-or-love:
Hi @awachtler! Great, thanks a lot for your proposal. I am not sure when exactly I will have a moment to give a review, but I will try to take a good look at it in the coming days. The functionality looks nice already :) 
In case of anything, @mplanchard, please feel free to take a second look

dee-me-tree-or-love:
Hi @awachtler! Sorry getting back to it only after such a delay, but I've taken a look and had to look a bit into Micropython to get slightly more familiar, but it looks like a nice addition in my opinion ;). I will add a few comments, but I think it'll make it nicely into master :). @mplanchard, if you'd have some time, could you give a second opinion?  

But to give a better idea, could you add an example of using this endpoint with `upip`? I think a nice spot for it would be in the readme as a new section in [`Recipes`](https://github.com/pypiserver/pypiserver#recipes)!

pypiserver/_app.py
```diff
@@ -362,6 +363,25 @@ def server_static(filename):
 
     return HTTPError(404, f"Not Found ({filename} does not exist)\n\n")
 
+@app.route("/:project/json")
+@auth("list")
+def json_info(project):
+    packages = sorted(
+        config.backend.find_project_packages(project),
```
dee-me-tree-or-love, line 370: I noticed that at `/:project` (which redirects to `/simple/:project`) pypiserver accepts only normalized project names according to [PEP 503](https://www.python.org/dev/peps/pep-0503/). Might be a detail, but I think it'd be good if we could keep this behaviour consistent. :) 

For instance, you could add a check similar to how it is implemented at [`_app.py:L279`](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/_app.py#L279).

```python
@app.route("/simple/:project/")
@auth("list")
def simple(project):
    # PEP 503: require normalized project
    normalized = normalize_pkgname_for_url(project)
    if project != normalized:
        return redirect(f"/simple/{normalized}/", 301)
    ...
```

pypiserver/_app.py
```diff
@@ -362,6 +363,25 @@ def server_static(filename):
 
     return HTTPError(404, f"Not Found ({filename} does not exist)\n\n")
 
+@app.route("/:project/json")
+@auth("list")
+def json_info(project):
```
dee-me-tree-or-love, line 368: [nit:] It would be grand (instant karma points) if you could add a unit test for this at `tests/test_app.py`.

dee-me-tree-or-love (changes requested):
Looks good, but I have left a couple of comments :)
1. could you align with PEP 503? 
2. if you could add a unit test, would be very grateful!

awachtler:
@dee-me-tree-or-love, I just added normalized package name redirecting. 

While testing I found that I should raise 404 error in case of invalid package name. 
Currently 3 tests are added, let me know, if more test coverage is required.

The use-case documentation for micropython I will add within the next days.

awachtler:
Just updated README.rst

README.rst
```diff
@@ -884,6 +884,32 @@ these steps:
...
+    import network
+    import upip
+
+    sta_if = network.WLAN(network.STA_IF)
+    sta_if.active(True)
+    sta_if.connect('<your ESSID>', '<your password>')
+    upip.index_urls = ["http://my-server:8080"]
+    upip.install("micropython-foobar")
+
+Furhter information on micropython-packaging can be found here: https://docs.micropython.org/en/latest/reference/packages.html
```
dee-me-tree-or-love, line 911:
```suggestion
Further information on micropython-packaging can be found here: https://docs.micropython.org/en/latest/reference/packages.html
```

pypiserver/_app.py
dee-me-tree-or-love: Perfect, thanks for addding!

dee-me-tree-or-love:
@awachtler great! Thanks a lot, I've seen the newly added changes! Looks great to me! :) 
I've spotted a small typo in the README and after that I'd be glad to approve :)

dee-me-tree-or-love (commented):
Spotted a typo ;)

awachtler:
Hallo, @dee-me-tree-or-love, thanks for your careful review, the typo was just fixed.

Issue #387, opened 2021-04-20
Add Healtcheck endpoint


rodrigo-fss:
It would be great if pypiserver had a health check endpoint.
something like /health, answering 200 OK
so it could be possible to serve this application using something like ECS or k8

dee-me-tree-or-love:
@rodrigo-fss I think it's a great idea. I will try try pick it up in the coming days. Otherwise feel free to create a MR if you'd like to add it yourself ;)

@mplanchard I think that's something small we could easily incorporate? What do you think?

PR #396, opened 2021-07-21
feat: add extremely basic /health endpoint

dee-me-tree-or-love:
# What

Following the request from #387 this MR adds an (extremely!) basic `/health` endpoint.  

Suggestions are welcome to make it into a more sophisticated solution :)


pypiserver/_app.py
```diff
@@ -81,6 +81,11 @@ def log_error(http_error):
     log.info(config.log_err_frmt, vars(http_error))
 
 
+@app.route("/health")
+def health():
+    return 'Ok.'
```
skarzi, line 86: Dot seems to be unnecessary:

```suggestion
    return 'Ok'
```

skarzi (commented):
First of all - thanks for developing this great project!

When do you plan to release a new version of `pypiserver` with this endpoint added?

pypiserver/_app.py
dee-me-tree-or-love: That's a good idea, indeed, the dot definitely can be skipped 👍  :)

dee-me-tree-or-love:
Hey, @skarzi thanks a lot for the review and the kind words! Really glad to hear.  
I applied the suggestion, it is definitely a good idea :)   

As for the new releases, I, unfortunately, don't have a release plan yet. Atm, I am spending less time on the project than I hoped for - doing my masters and having a job does have an impact on my availability :) really sorry about this.  
But it's been on my mind for a while now, I'd like to set up an automated release/deployment schedule so that it can be as autonomous as possible and we'll start getting the latest features back in. So thanks for the reminder!   
I will make an issue with this idea in any case in the coming days 👍 :) But getting to work on it, I think I will be able only in a week or two at the earliest, but I will keep you posted :)

--- 

Btw, if you are interested, we've got an open call for new contributors, so will be very happy to see anyone who'd like to get helping to hop on board :D.

skarzi:
@dee-me-tree-or-love if you need any help with GitHub Actions workflows for auto-deployment or release - I can help with it, let's just discuss how it should work in some issue

dee-me-tree-or-love:
> @dee-me-tree-or-love if you need any help with GitHub Actions workflows for auto-deployment or release - I can help with it, let's just discuss how it should work in some issue

hey @skarzi, thanks a lot! Really appreciate the offer, that'd be great. I have outlined a few thoughts on the topic in the following issues #405 and #404, and would really appreciate some suggestions 🙌 Trying my best here, but time-wise, I don't have a very clear decision in mind.

PR #403, opened 2021-10-20
Switch to a PollingObserver from watchdog if any root is NFS

### LINKED ISSUE #324: Cache - not working
- running pypiserver with cache in a container
- volume mounted an NFS directory where the package wheel files (package_root)

we ended up in a situation where the files we put into the directory was not being seen - due to cache. 

the same hosting system without the cache works - all files are seen, albeit a bit slower.

g-m-o-n-e-y:
Fixes #324 

First time posting a fix so let me know if I'm missing anything.

We're running into this issue in a k8s cluster where our IT dept only allows NFS shares for persistent storage. We notice that  uploading to one pod with this shared storage doesn't update the other pods.

Also note that I'm punting on Windows as it appears much easier to tell if the folder is NFS on *nix.

### DIFF
--- a/pypiserver/cache.py
+++ b/pypiserver/cache.py
@@ -3,19 +3,23 @@
 # is installed
 #
 
+import platform
 from os.path import dirname
 from pathlib import Path
+from subprocess import call
 import typing as t
 import threading
 
 try:
     from watchdog.observers import Observer
+    from watchdog.observers.polling import PollingObserver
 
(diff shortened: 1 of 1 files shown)


dee-me-tree-or-love:
Hey @g-m-o-n-e-y, thanks for the pull request! I will try to get to this very soon. A little bit busy these days, but I will keep you posted when I manage to take a look 👍

Issue #406, opened 2021-11-16
Tests failing for tox targeting Python 3.9. 


dee-me-tree-or-love:
As noticed in #396, the latest attempts to run `test-cpython3.9` GitHub Action fails, meanwhile no code changes could trigger the error. This seems to be related to updated dependencies of `tox` as installed during the `Install dependencies` step in `test-cpython3.9`. 

[Currently passing commit on `master` branch](https://github.com/pypiserver/pypiserver/runs/3767104332?check_suite_focus=true) has the following dependencies installed:  
```bash
Successfully installed:
backports.entry-points-selectable-1.1.0
distlib-0.3.3
filelock-3.2.0
packaging-21.0
platformdirs-2.4.0
pluggy-1.0.0
py-1.10.0
pyparsing-2.4.7
six-1.16.0
toml-0.10.2
tox-3.24.4
virtualenv-20.8.1
```

While the [failing run on the same commit on a fork of `master` branch](https://github.com/dee-me-tree-or-love/pypiserver/runs/4227358183?check_suite_focus=true) has the following installations: 
```bash
Successfully installed:
backports.entry-points-selectable-1.1.1 # ( backports.entry-points-selectable-1.1.0 in previous run! )
distlib-0.3.3
filelock-3.4.0 # ( filelock-3.2.0 in previous run! )
packaging-21.2 # ( packaging-21.0 in previous run! )
platformdirs-2.4.0
pluggy-1.0.0
py-1.11.0 # ( py-1.11.0 in previous run! )
pyparsing-2.4.7
six-1.16.0
toml-0.10.2
tox-3.24.4
virtualenv-20.10.0 # ( virtualenv-20.8.1 in previous run! )
```

My suspicion is that the failures are related to one of the package updates. Any additional insights are welcome. 

## Proposed actions: 

- [ ] Pin `tox` dependencies to the working version
- [ ] Investigate which of the changes cause the issue

dee-me-tree-or-love:
As a temporary step to resolve the CI/CD errors and unblock the pull requests, disabling the tests for 3.9 with #407.

PR #407, opened 2021-11-16
Temporary disable tests for python3.9

dee-me-tree-or-love:
As a temporary solution for #406 to unblock the CI/CD on other pull requests.


dee-me-tree-or-love:
Noticed some strange behaviour with `test-pypy` unpredictably failing to run `test_server.py` with timeouts. Will address it here #408.

Issue #439, opened 2022-08-22
Failing `test-docker` CI step in github actions workflow


dee-me-tree-or-love:
# What

The `test-docker` step of the `ci` workflow seems to fail on installing `gevent=21.1.2` (here are the [job logs](https://github.com/pypiserver/pypiserver/runs/7949072664?check_suite_focus=true#step:6:307)) and some other issues. This prevents the CI pipeline from completing successfully and should be resolved.

Issue #368, opened 2021-02-02
Update contributor docs


mplanchard:
Contributor docs definitely could use some expansion. I'd like to add at least:

- [ ] Architecture overview
- [ ] Any standards and style guidelines we can't enforce with linting
- [ ] Expectations for tests
- [ ] Expectations for backwards compatibility 
- [ ] Guidelines for issue triage
- [ ] Deploy process

----

- [ ] TODO add this to the v2 milestone when I'm not on mobile

dee-me-tree-or-love:
I can pick it up

mplanchard:
Awesome! Feel free to pick up any part of it you feel comfortable with. Some of this stuff doesn't exist anywhere but in my head at this point, so also feel free to hit me up if you're blocked

dee-me-tree-or-love:
@mplanchard perfect, yes! I will pick up the architecture overview and maybe the deploy process as first candidates. But will take a closer look when picking it up these days :) I will poke you on Zulip if anything further.

dee-me-tree-or-love:
Will be adding documentation in this branch on my fork https://github.com/dee-me-tree-or-love/pypiserver/tree/368_update-contributor-docs_first-setup :v: any suggestions are welcome

dee-me-tree-or-love:
Adding another documentation area: 
- [ ] document the release management flow https://github.com/pypiserver/pypiserver/issues/405

dee-me-tree-or-love:
Also related PR on documentation #451

PR #442, opened 2022-08-28
feat: :stethoscope: allow customized health check endpoint

Mason-Lin:
~~Get the liveness endpoint from the environment variable `HEALTH_ENDPOINT` and verify it. If the customized endpoint is invalied, it will fallback to the DEFAULT_HEALTH_ENDPOINT.~~

UPDATE 2022-10-21
----------------------------

`pypiserver` provides a default health endpoint at `/health`. It always returns
`200 Ok` if the service is up. Otherwise, it means that the service is not responsive.

In addition, `pypiserver` allows users to customize the health endpoint.

Alphanumeric characters, hyphens, forward slashes and underscores are allowed and the endpoint should not overlap with any existing routes.

Valid examples: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -920,8 +920,8 @@ Further information on micropython-packaging can be found here: https://docs.mic
 Custom Health Check Endpoint
 ----------------------------
 
-pypiserver provide a default health endpoint. ``/health`` It always returns
-``200 Ok`` if the service is up. Otherwise it means that the service is dead.
+``pypiserver`` provides a default health endpoint at ``/health``. It always returns
+``200 Ok`` if the service is up. Otherwise, it means that the service is not responsive.
 
 pypiserver also allow users to customize the health endpoint. (see `#441 <https://github.com/pypiserver/pypiserver/issues/441>`_).
 Alphanumeric characters, hyphens, forward slashes and underscores are allowed.
--- a/README.rst
+++ b/README.rst
@@ -945,7 +945,7 @@ Configure a custom health endpoint by script
     app = pypiserver.app(root="./packages", health_endpoint="/action/health")
     bottle.run(app=app, host="0.0.0.0", port=8080, server="auto")
 
-Try `curl http://localhost:8080/action/health`
+Try ``curl http://localhost:8080/action/health``
 
 
 Sources
--- a/README.rst
+++ b/README.rst
@@ -135,6 +135,10 @@ not officially supported, and will not receive bugfixes or new features.
         For packages not found in the local index, this URL will be used to
         redirect to (default: https://pypi.org/simple/).
 
+      --health-endpoint HEALTH_ENDPOINT
+        Configure a custom liveness endpoint. It always returns 200 Ok if 
+        the service is up. Otherwise it means that the service is dead.
+
       --server METHOD
         Use METHOD to run the server. Valid values include paste,
         cherrypy, twisted, gunicorn, gevent, wsgiref, auto. The
--- a/README.rst
+++ b/README.rst
@@ -920,18 +920,17 @@ Further information on micropython-packaging can be found here: https://docs.mic
 Custom Health Check Endpoint
 ----------------------------
 
-pypiserver provide a default health endpoint. `/health`
-It always returns 200 Ok if the service is up. Otherwise it means that the 
-service is dead.
+pypiserver provide a default health endpoint. ``/health`` It always returns
+``200 Ok`` if the service is up. Otherwise it means that the service is dead.
 
 pypiserver also allow users to customize the health endpoint. (see `#441 <https://github.com/pypiserver/pypiserver/issues/441>`_).
 Alphanumeric characters, hyphens, forward slashes and underscores are allowed.
-Such as: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`
+Valid example: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`
 
 Configure a custom health endpoint by CLI arguments
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Run pypiserver with argument::
+Run pypiserver with ``--health-endpoint`` argument::
 
     pypi-server --health-endpoint /action/health
 
--- a/README.rst
+++ b/README.rst
@@ -917,6 +917,36 @@ Installing packages from the REPL of an embedded device works in this way:
 
 Further information on micropython-packaging can be found here: https://docs.micropython.org/en/latest/reference/packages.html
 
+Custom Health Check Endpoint
+----------------------------
+
+pypiserver provide a default health endpoint. `/health`
+It always returns 200 Ok if the service is up. Otherwise it means that the 
+service is dead.
+
+pypiserver also allow users to customize the health endpoint. (see `#441 <https://github.com/pypiserver/pypiserver/issues/441>`_).
+Alphanumeric characters, hyphens, forward slashes and underscores are allowed.
+Such as: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`
+
+Configure a custom health endpoint by CLI arguments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Run pypiserver with argument::
+
+    pypi-server --health-endpoint /action/health
+
+Configure a custom health endpoint by script
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    import pypiserver
+    from pypiserver import bottle
+    app = pypiserver.app(root="./packages", health_endpoint="/action/health")
+    bottle.run(app=app, host="0.0.0.0", port=8080, server="auto")
+
+Try `curl http://localhost:8080/action/health`
+
 
 Sources
 =======
--- a/README.rst
+++ b/README.rst
@@ -136,8 +136,8 @@ not officially supported, and will not receive bugfixes or new features.
         redirect to (default: https://pypi.org/simple/).
 
       --health-endpoint HEALTH_ENDPOINT
-        Configure a custom liveness endpoint. It always returns 200 Ok if 
-        the service is up. Otherwise it means that the service is dead.
+        Configure a custom liveness endpoint. It always returns 200 Ok if  
+        the service is up. Otherwise, it means that the service is not responsive.
 
       --server METHOD
         Use METHOD to run the server. Valid values include paste,
--- a/README.rst
+++ b/README.rst
@@ -923,9 +923,10 @@ Custom Health Check Endpoint
 ``pypiserver`` provides a default health endpoint at ``/health``. It always returns
 ``200 Ok`` if the service is up. Otherwise, it means that the service is not responsive.
 
-pypiserver also allow users to customize the health endpoint. (see `#441 <https://github.com/pypiserver/pypiserver/issues/441>`_).
-Alphanumeric characters, hyphens, forward slashes and underscores are allowed.
-Valid example: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`
+In addition, ``pypiserver`` allows users to customize the health endpoint.
+Alphanumeric characters, hyphens, forward slashes and underscores are allowed
+and the endpoint should not overlap with any existing routes.
+Valid examples: ``/healthz``, ``/health/live-1``, ``/api_health``, ``/action/health``
 
 Configure a custom health endpoint by CLI arguments
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -147,11 +147,20 @@ def app_from_config(config: RunConfig) -> Bottle:
     # Add a reference to our config on the Bottle app for easy access in testing
     # and other contexts.
     _app.app._pypiserver_config = config
-    # add customized health check endpoint
-    _app.app.route(_get_health_endpoint(config, _app.app), "GET", lambda: "Ok")
     return _app.app
 
 
+def setup_routes_from_config(app: Bottle, config: RunConfig) -> Bottle:
+    """Set up additional routes supplied from the config."""
+    def _setup_health_endpoint(app, config):
+        if config.health_endpoint in [route.rule for route in app.routes]:
+            raise RuntimeError(
+                "Provided health endpoint overlaps with existing routes"
+            )
+        app.route(config.health_endpoint, "GET", lambda: "Ok")
+
+    _setup_health_endpoint(app, config)
+    return app
 T = t.TypeVar("T")
 
 
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -5,7 +5,7 @@ import sys
 import typing as t
 
 from pypiserver.bottle import Bottle
-from pypiserver.config import Config, RunConfig, strtobool
+from pypiserver.config import DEFAULTS, Config, RunConfig, strtobool
 
 version = __version__ = "1.5.0"
 __version_info__ = tuple(_re.split("[.-]", __version__))
@@ -114,6 +114,15 @@ def backwards_compat_kwargs(kwargs: dict, warn: bool = True) -> dict:
     return updated_kwargs
 
 
+def _get_health_endpoint(config: RunConfig, app: Bottle) -> str:
+    """Verify the health_endpoint and fallback to default if invalid."""
+    # Avoid health check route uses exist endpoints
+    for rule in (route.rule for route in app.routes if route.rule != "/"):
+        if config.health_endpoint.startswith(rule):
+            return DEFAULTS.HEALTH_ENDPOINT
+    return config.health_endpoint
+
+
 def app(**kwargs: t.Any) -> Bottle:
     """Construct a bottle app running pypiserver.
 
@@ -138,6 +147,8 @@ def app_from_config(config: RunConfig) -> Bottle:
     # Add a reference to our config on the Bottle app for easy access in testing
     # and other contexts.
     _app.app._pypiserver_config = config
+    # add customized health check endpoint
+    _app.app.route(_get_health_endpoint(config, _app.app), "GET", lambda: "Ok")
     return _app.app
 
 
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -143,6 +143,7 @@ def app_from_config(config: RunConfig) -> Bottle:
 
 def setup_routes_from_config(app: Bottle, config: RunConfig) -> Bottle:
     """Set up additional routes supplied from the config."""
+
     def _setup_health_endpoint(app, config):
         if config.health_endpoint in [route.rule for route in app.routes]:
             raise RuntimeError(
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -5,7 +5,7 @@ import sys
 import typing as t
 
 from pypiserver.bottle import Bottle
-from pypiserver.config import DEFAULTS, Config, RunConfig, strtobool
+from pypiserver.config import Config, RunConfig, strtobool
 
 version = __version__ = "1.5.0"
 __version_info__ = tuple(_re.split("[.-]", __version__))
--- a/pypiserver/__init__.py
+++ b/pypiserver/__init__.py
@@ -114,15 +114,6 @@ def backwards_compat_kwargs(kwargs: dict, warn: bool = True) -> dict:
     return updated_kwargs
 
 
-def _get_health_endpoint(config: RunConfig, app: Bottle) -> str:
-    """Verify the health_endpoint and fallback to default if invalid."""
-    # Avoid health check route uses exist endpoints
-    for rule in (route.rule for route in app.routes if route.rule != "/"):
-        if config.health_endpoint.startswith(rule):
-            return DEFAULTS.HEALTH_ENDPOINT
-    return config.health_endpoint
-
-
 def app(**kwargs: t.Any) -> Bottle:
     """Construct a bottle app running pypiserver.
 
@@ -130,7 +121,7 @@ def app(**kwargs: t.Any) -> Bottle:
         (or its base), defined in `pypiserver.config`, may be overridden.
     """
     config = Config.default_with_overrides(**backwards_compat_kwargs(kwargs))
-    return app_from_config(config)
+    return setup_routes_from_config(app_from_config(config), config)
 
 
 def app_from_config(config: RunConfig) -> Bottle:
@@ -161,6 +152,8 @@ def setup_routes_from_config(app: Bottle, config: RunConfig) -> Bottle:
 
     _setup_health_endpoint(app, config)
     return app
+
+
 T = t.TypeVar("T")
 
 
--- a/pypiserver/config.py
+++ b/pypiserver/config.py
@@ -402,7 +402,7 @@ def get_parser() -> argparse.ArgumentParser:
         type=health_endpoint_arg,
         help=(
             "Configure a custom liveness endpoint. It always returns 200 Ok if "
-            "the service is up. Otherwise it means that the service is dead."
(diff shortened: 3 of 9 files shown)


Mason-Lin:
Hi @dee-me-tree-or-love,
May I get your approval to run the tests workflow?

dee-me-tree-or-love:
Hey @Mason-Lin sure! The CI's on for the PR :) Thanks for sending in the idea! I will try to leave a proper review a little later today. 👍

Mason-Lin:
> Hey @Mason-Lin sure! The CI's on for the PR :) Thanks for sending in the idea! I will try to leave a proper review a little later today. 👍

Thank you so much!

Mason-Lin:
Hi @dee-me-tree-or-love ,

In the beginning, I tried to copy FALLBACK_URL and add --health-endpoint to [DEFAULTS](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/config.py#L78) and [RunConfig](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/config.py#L660) and [run_parser](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/config.py#L377)

Then I replaced [`@app.route("/health")`](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/_app.py#L85) with `@app.route(config.health_endpoint)` 

I got an error while executing the test

Then I checked [`app = pypiserver.app_from_config(config)`](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/__main__.py#L171)
I found that [`app_from_config`](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/__init__.py#L127) is use `__import__` to `import _app` first, and then mutate global variable `_app.config = config`

Because the `bottle.route` decorator already needs `config.health_endpoint` when `import _app`, it seems that the current import order cannot get CLI arguments while creating the bottle instance.
I'm not quite sure how to solve this.
I saw a function in the bottle [`add_route`](https://github.com/pypiserver/pypiserver/blob/master/pypiserver/bottle.py#L775), it may be able to delay the time of adding endpoints but I am not very familiar with the bottle.

The above is the blockage I encountered while exploring the CLI interface.😅

I would greatly appreciate it if you kindly give me some suggestions.

dee-me-tree-or-love:
Hey @Mason-Lin! Thanks a lot for the detailed description, this helps a lot, indeed I see the trouble 👍 I'm sorry I didn't get back to this yesterday, had some unexpected changes to my plans, but I will try to think along in the coming days and will keep you posted. 

What you describe is indeed a bit of a blocking issue, so I will try to see how to work around this with minimal changes. 

Alternatively, I have come across the _explicit routing_ of the bottle library: (see [Request Routing -> EXPLICIT ROUTING CONFIGURATION](https://bottlepy.org/docs/dev/routing.html#explicit-routing-configuration)). This seems to allow defining the routes within a function, which could technically solve the import order issue. 

```python
# Taken from the bottle docs:
#  see https://bottlepy.org/docs/dev/routing.html#explicit-routing-configuration

def setup_routing(app):
    app.route('/new', ['GET', 'POST'], form_new)
    app.route('/edit', ['GET', 'POST'], form_edit)

app = Bottle()
setup_routing(app)
```

The only thing is that I am a bit wary of coupling our code to the bottle's interface even more. And if that is to be the solution, I'd like to pick this up in a separate PR to keep the changes isolated. 

But feel to give it a try, maybe that makes it work with `config`, and I will also try to get myself some time to try it out in code too :) 

Furthermore, not to make many changes to the internals of `pypiserver`, it could be a compromise to use the explicit routing approach, but on the client side, when configuring your `pypiserver` instance. A little similar to what I did with adding hooks after the app creation in a separate use case ([this `get_app` on the client side](https://github.com/dee-me-tree-or-love/pypiserver-appengine/blob/master/main.py#L399)). This could also be a way to go around this, making all the changes purely on the client side :) 

In any case, I'll get back to this soon, thanks a lot for all your work! 👍

Mason-Lin:
Hi @dee-me-tree-or-love ,

Thank you for the suggestions.
I'll find time next week to look into these solutions. 😀

Mason-Lin:
Thanks again for the suggestion
Using _explicit routing_ can solve the problem I had before 👍

The practice of `get_app()` mentioned to set `entrypoint: gunicorn -b :8081 -w 2 'main:get_app()'`
I haven't experimented with this method yet, but it seems that adding a health endpoint can be done without modifying `pypiserver`.
Once get the `app` from pypiserver
The user should be able to patch their endpoint from the client side.
Making all the changes purely on the client side.

My current usage is actually to deploy `pypiserver/pypiserver:latest` with docker
Then start a private PyPI server in the private network through CMD ["run", "--disable-fallback"]
I wish I could keep using `entrypoint.sh` and passing some more parameters to accomplish the requirement.
I hope the customized health endpoint becomes a feature out of the box.

I will now change the environment variable approach to using the CLI to set the health endpoint. 😊

Mason-Lin:
notes:
build fail related to #439

dee-me-tree-or-love:
Hey, @Mason-Lin I've fixed the #439 issue with #444. The CI should be back to normal if you update from `master`. I will get to reviewing this more properly soon, thanks for all the updates! :)

dee-me-tree-or-love:
Hey @Mason-Lin, thanks a lot for all the updates on this branch! :v: I've had a few things that kept me busy last week, but I'll try to walk through your PR in detail this week :D

Mason-Lin:
> Hey @Mason-Lin, thanks a lot for all the updates on this branch! :v: I've had a few things that kept me busy last week, but I'll try to walk through your PR in detail this week :D 

No problem, thank you 😁

pypiserver/config.py
```diff
@@ -382,6 +391,15 @@ def get_parser() -> argparse.ArgumentParser:
             "index."
         ),
     )
+    run_parser.add_argument(
+        "--health-endpoint",
+        default=DEFAULTS.HEALTH_ENDPOINT,
+        type=health_endpoint_arg,
+        help=(
+            "The liveness endpoint. Always response 200 OK, otherwise means"
+            "that the service is unhealthy or dead."
```
dee-me-tree-or-love, line 400:
```suggestion
            "Configure a custom liveness endpoint. It always returns 200 OK if the service is up." 
            "Otherwise it means that the service is unhealthy or dead."
```

I think this makes the purpose of this setting a little more clear :)

tests/test_app.py
```diff
@@ -188,7 +188,27 @@ def test_packages_empty(testapp):
...
+
+def test_health_customized_endpoint(root):
+    from pypiserver import app
+
+    testapp = webtest.TestApp(
+        app(root=root.strpath, health_endpoint="/healthz")
+    )
+    resp = testapp.get("/healthz")
+    assert resp.status_int == 200
+    assert "Ok" in resp
```
dee-me-tree-or-love, line 205: Hey, these are really nice tests, thanks a lot for adding! 🤩

tests/test_app.py
```diff
@@ -188,7 +188,27 @@ def test_packages_empty(testapp):
...
+
+
+def test_health_customized_endpoint(root):
+    from pypiserver import app
+
+    testapp = webtest.TestApp(
+        app(root=root.strpath, health_endpoint="/healthz")
+    )
```
dee-me-tree-or-love, line 202:
```suggestion
    _app = app(root=root.strpath, health_endpoint="/healthz")
    testapp = webtest.TestApp(_app)
```

Just to keep the lines similar to other test setups :)

tests/test_app.py
```diff
@@ -188,7 +188,27 @@ def test_packages_empty(testapp):
...
+    resp = testapp.get("/healthz")
+    assert resp.status_int == 200
+    assert "Ok" in resp
+
+
+def test_health_invalid_customized_endpoint(root):
+    from pypiserver import app
+
+    testapp = webtest.TestApp(app(root=root.strpath, health_endpoint="/simple"))
```
dee-me-tree-or-love, line 211: Also linked to some comments above, I'd say that here it should raise a `RuntimeError` since this route is already taken :)

tests/test_config.py
```diff
@@ -369,6 +369,28 @@ def generate_subcommand_test_cases(
...
+        legacy_args=["--health-endpoint", "/healthz"],
+        exp_config_type=RunConfig,
+        exp_config_values={"health_endpoint": "/healthz"},
+    ),
+    ConfigTestCase(
+        case="Run: health-endpoint specified but invalid",
+        args=["run", "--health-endpoint", "/health!"],
+        legacy_args=["--health-endpoint", "/health!"],
+        exp_config_type=RunConfig,
+        exp_config_values={"health_endpoint": DEFAULTS.HEALTH_ENDPOINT},
```
dee-me-tree-or-love, line 392: This case should raise an error, similar to the `case=f"Invalid hash algo: {val}"`, as this would be a more straightforward outcome imho :) Linked to a comment above on `health_endpoint_arg`. And for inspiration on how to set the error cases up, see `_CONFIG_ERROR_CASES`.

pypiserver/__init__.py
```diff
@@ -5,7 +5,7 @@
 import typing as t
 
 from pypiserver.bottle import Bottle
-from pypiserver.config import Config, RunConfig, strtobool
+from pypiserver.config import DEFAULTS, Config, RunConfig, strtobool
```
dee-me-tree-or-love, line 8: I think we should keep the `DEFAULTS` as much as possible to the `config.py` module, so as to prevent the mixture of responsibilities :) I elaborate on it in other comments.

```suggestion
from pypiserver.config import Config, RunConfig, strtobool
```

dee-me-tree-or-love (changes requested):
Hey @Mason-Lin, thanks a lot for your work and adjusting it to the explicit routing setup! 😊 This is already looking nicer, and I think this addition could be an interesting enabler for some extra configurations for the `pypiserver` in the future. I added a few comments with requests to change below. Those are some alignment details to keep this new behaviour as close as possible to other features of `pypiserver`, plus to make sure it's modular enough for future extensions or changes. I think it's getting there! Could you take a look at the comments? :) 

Special thanks for making all the test cases, these are useful and helpful to see! Then, when you finish this up, don't forget to add a note about this in the `Recipes` section of the `README` and update the `help` command output in the `Quickstart` section, this will be nice ;) 👍 🚀

pypiserver/config.py
dee-me-tree-or-love: The line length in this suggestion might need to be double checked 😁

Mason-Lin:
> Hey @Mason-Lin, thanks a lot for your work and adjusting it to the explicit routing setup! 😊 This is already looking nicer, and I think this addition could be an interesting enabler for some extra configurations for the `pypiserver` in the future. I added a few comments with requests to change below. Those are some alignment details to keep this new behaviour as close as possible to other features of `pypiserver`, plus to make sure it's modular enough for future extensions or changes. I think it's getting there! Could you take a look at the comments? :) 
> 
> 
> 
> Special thanks for making all the test cases, these are useful and helpful to see! Then, when you finish this up, don't forget to add a note about this in the `Recipes` section of the `README` and update the `help` command output in the `Quickstart` section, this will be nice ;) 👍 🚀  

Hi @dee-me-tree-or-love,

Ok no problem, thank you very much for your review.
I have a work in progress that will probably be done this week, and then I'll start working on this PR 😁

tests/test_app.py
Mason-Lin: I think you have a point, Explicit is better than implicit.
https://github.com/pypiserver/pypiserver/pull/442/commits/55fa396999c07be76ed58cfe4b14d860809368a1

pypiserver/__init__.py
```diff
@@ -114,6 +114,15 @@ def backwards_compat_kwargs(kwargs: dict, warn: bool = True) -> dict:
     return updated_kwargs
 
 
+def _get_health_endpoint(config: RunConfig, app: Bottle) -> str:
+    """Verify the health_endpoint and fallback to default if invalid."""
+    # Avoid health check route uses exist endpoints
+    for rule in (route.rule for route in app.routes if route.rule != "/"):
+        if config.health_endpoint.startswith(rule):
+            return DEFAULTS.HEALTH_ENDPOINT
```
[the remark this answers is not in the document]
Mason-Lin: That's a suggestion👍

When route conflicts it will [raise RuntimeError](https://github.com/pypiserver/pypiserver/pull/442/commits/55fa396999c07be76ed58cfe4b14d860809368a1), so the [DEFAULTS is no longer needed](https://github.com/pypiserver/pypiserver/pull/442/commits/d2849d8a48f499628d80d7e015bf6e119cbe7220) during [setup_routes_from_config()](https://github.com/pypiserver/pypiserver/pull/442/commits/f1ea6707739e4720c93430df114926e11c03a731)

tests/test_config.py
Mason-Lin: [raise SystemExit when using invalid args](https://github.com/pypiserver/pypiserver/pull/442/commits/edcf5778e732440f506b1ec1222fe7d78d0fc1c1)

pypiserver/config.py
```diff
@@ -134,6 +135,14 @@ def hash_algo_arg(arg: str) -> t.Optional[str]:
     )
 
 
+def health_endpoint_arg(arg: str) -> str:
+    """Verify the health_endpoint and fallback to default if invalid."""
+    # Only allow alphanumeric, hyphen, forward slash, and underscore
+    if re.fullmatch(r"^/[a-z0-9/_-]+$", arg, re.I) is None:
+        return DEFAULTS.HEALTH_ENDPOINT
+    return arg
```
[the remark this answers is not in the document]
Mason-Lin: [Config: raise SystemExit when using invalid args](https://github.com/pypiserver/pypiserver/pull/442/commits/35df0cfb8fdbd6ec1733788d503dd9117ba90ccd)

Mason-Lin:
Hi @dee-me-tree-or-love ,
Thank you for your good advice and clear explanation of ideas considering project maintenance and user experience :)

Mason-Lin:
Hi @dee-me-tree-or-love ,
Could you enabled the GitHub actions to run on this PR again? so I can fix them before next review. 🫡

README.rst
```diff
@@ -913,6 +917,35 @@ Installing packages from the REPL of an embedded device works in this way:
 
 Further information on micropython-packaging can be found here: https://docs.micropython.org/en/latest/reference/packages.html
 
+Custom Health Check Endpoint
+----------------------------
+
+pypiserver provide a default health endpoint. ``/health`` It always returns
+``200 Ok`` if the service is up. Otherwise it means that the service is dead.
```
dee-me-tree-or-love, line 924:
```suggestion
``pypiserver`` provides a default health endpoint at ``/health``. It always returns
``200 Ok`` if the service is up. Otherwise, it means that the service is not responsive.
```

Some small wording suggestions :)

README.rst
```diff
@@ -913,6 +917,35 @@ Installing packages from the REPL of an embedded device works in this way:
...
+
+pypiserver also allow users to customize the health endpoint. (see `#441 <https://github.com/pypiserver/pypiserver/issues/441>`_).
+Alphanumeric characters, hyphens, forward slashes and underscores are allowed.
+Valid example: `/healthz`, `/health/live-1`, `/api_health`, `/action/health`
```
dee-me-tree-or-love, line 928:
```suggestion
In addition, ``pypiserver`` allows users to customize the health endpoint.
Alphanumeric characters, hyphens, forward slashes and underscores are allowed
and the endpoint should not overlap with any existing routes.
Valid examples: ``/healthz``, ``/health/live-1``, ``/api_health``, ``/action/health``
```

Some more small suggestions :)

README.rst
```diff
@@ -913,6 +917,35 @@ Installing packages from the REPL of an embedded device works in this way:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    import pypiserver
+    from pypiserver import bottle
+    app = pypiserver.app(root="./packages", health_endpoint="/action/health")
+    bottle.run(app=app, host="0.0.0.0", port=8080, server="auto")
+
+Try `curl http://localhost:8080/action/health`
```
dee-me-tree-or-love, line 947:
```suggestion
Try ``curl http://localhost:8080/action/health``
```

Just for consistency, otherwise, RST doesn't render this as code blocks :)

tests/test_config.py
```diff
@@ -725,7 +748,7 @@ def test_config(
 @pytest.mark.parametrize(
     "args, exp_txt",
     CONFIG_ERROR_PARAMS,
-    ids=CONFIG_TEST_IDS,
+    ids=CONFIG_ERROR_IDS,
```
dee-me-tree-or-love, line 751: 🤩  great catch! Thanks 👍

pypiserver/config.py
```diff
@@ -382,6 +396,15 @@ def get_parser() -> argparse.ArgumentParser:
             "index."
         ),
     )
+    run_parser.add_argument(
+        "--health-endpoint",
+        default=DEFAULTS.HEALTH_ENDPOINT,
+        type=health_endpoint_arg,
+        help=(
+            "Configure a custom liveness endpoint. It always returns 200 Ok if "
+            "the service is up. Otherwise it means that the service is dead."
```
dee-me-tree-or-love, line 405:
```suggestion
            "the service is up. Otherwise, it means that the service is not responsive."
```

Just as an alignment :)

tests/test_main.py
```diff
@@ -294,3 +294,20 @@ def test_auto_servers() -> None:
...
+
+def test_health_endpoint_customized(main):
+    main(["--health-endpoint", "/healthz"])
+    assert main.app._pypiserver_config.health_endpoint == "/healthz"
+    assert "/healthz" in (route.rule for route in main.app.routes)
+
+
+def test_health_endpoint_invalid_customized(main):
+    with pytest.raises(SystemExit):
+        main(["--health-endpoint", "/health!"])
```
dee-me-tree-or-love, line 313: Awesome! Thanks a lot for setting up all these tests!

dee-me-tree-or-love (dismissed):
Hey @Mason-Lin, this is great work! Thanks a lot for the effort, it was a pleasure to discuss this change 👍 Sorry for my late response, last few weeks I had to manage some challenges on the personal side and was largely AFK. But now getting back to this :) I left a few final nitty comments to polish up the readme and the rest, the code looks good to go IMHO 👍

pypiserver/__init__.py
```diff
@@ -141,6 +141,19 @@ def app_from_config(config: RunConfig) -> Bottle:
     return _app.app
 
 
+def setup_routes_from_config(app: Bottle, config: RunConfig) -> Bottle:
+    """Set up additional routes supplied from the config."""
+    def _setup_health_endpoint(app, config):
```
dee-me-tree-or-love, line 146: I see the `black` checks are failing here asking for an extra line. 
image

You can run the autoformatting on all files using `black .` I think.

tests/test_main.py
Mason-Lin: I'm a fan of TDD 😄

pypiserver/__init__.py
Mason-Lin: Thank you, I'm sorry for making mistakes with formatting and typing over and over again 😅
I'm a little unclear on what tests should be done on the local side and how.
I later tried to build `venv`, pip installed `dev.pip`, and executed the `check` related command in `ci.yml` step by step
I hope it won't build fail this time

Mason-Lin:
Thanks for the correction in reStructuredText.
I usually use Markdown, I am unfamiliar with reStructuredText, and I misuse syntax ^^"

Mason-Lin:
> Hey @Mason-Lin, this is great work! Thanks a lot for the effort, it was a pleasure to discuss this change 👍 Sorry for my late response, last few weeks I had to manage some challenges on the personal side and was largely AFK. But now getting back to this :) I left a few final nitty comments to polish up the readme and the rest, the code looks good to go IMHO 👍

no problem, thanks for your time 👍

Mason-Lin:
Hi @dee-me-tree-or-love ,
  Do I need to update the branch or keep what it is?
  If it needs to be updated, should I rebase it or merge it?

Thank you.

pypiserver/__init__.py
dee-me-tree-or-love: Hey, no worries! I'm happy it helped! :)) yes, that's a good point actually, thanks for pointing this out, I think it'll be a nice improvement to contributor docs in general (#368) 👍 Now all looks green, great job 😃

README.rst
```diff
@@ -135,6 +135,10 @@ not officially supported, and will not receive bugfixes or new features.
         For packages not found in the local index, this URL will be used to
         redirect to (default: https://pypi.org/simple/).
 
+      --health-endpoint HEALTH_ENDPOINT
+        Configure a custom liveness endpoint. It always returns 200 Ok if 
+        the service is up. Otherwise it means that the service is dead.
```
dee-me-tree-or-love, line 140:
```suggestion
        Configure a custom liveness endpoint. It always returns 200 Ok if  
        the service is up. Otherwise, it means that the service is not responsive.
```

Just this tiny detail for consistency, otherwise it's great to go! :) Sorry for taking such a long time for the review 😅

dee-me-tree-or-love (dismissed):
Hey @Mason-Lin! Great job, I think this is ready to go, sorry for taking such a long time for this process, but I'll be happy to merge this :) There is one last tiny remark in the code, and regarding the branch update, if you could, that'd indeed be perfect 🙏  

> Hi @dee-me-tree-or-love ,
> Do I need to update the branch or keep what it is?
> If it needs to be updated, should I rebase it or merge it?
>
> Thank you.

You can do either a rebase or merge, there's no strict process on this since we're squashing PRs into the main branch in the end, it will all appear as a single commit borrowing the PR title :)

Mason-Lin:
Don't worry, I've also learned some good ideas along the way, thanks for every suggestion you give.

Regarding the workflow part, because “First-time contributors need a maintainer to approve running workflows”, I have an idea but I don't know if it is suitable or not, that is to provide a dockerfile, which contains mypy, pylint, black, etc.

If I want to confirm whether the PR will be pass the workflow, I can check with docker build.

dee-me-tree-or-love (approved):
Great work! 👍 :) Merging

dee-me-tree-or-love:
> Regarding the workflow part, because “First-time contributors need a maintainer to approve running workflows”, I have an idea but I don't know if it is suitable or not, that is to provide a dockerfile, which contains mypy, pylint, black, etc.
> 
> If I want to confirm whether the PR will be pass the workflow, I can check with docker build.

Hey @Mason-Lin, I like this idea! I totally agree that some improved flow for checking will be nice :) Maybe a tool like [`pre-commit`](https://pre-commit.com/), [`drkns`](https://github.com/frantzmiccoli/drkns), or even a collection of scripts in a `Makefile` may be less resource-heavy but allow the same as Dockerfile for checking the code styles :) However the Dockerfile will ensure that the environment is consistent and of course, in that sense, it would also be very beneficial 👍 Maybe a combination of these options will be a nice middle ground, thanks for the suggestion, I really like the proposal.

Mason-Lin:
> Hey @Mason-Lin, I like this idea! I totally agree that some improved flow for checking will be nice :) Maybe a tool like [`pre-commit`](https://pre-commit.com/), [`drkns`](https://github.com/frantzmiccoli/drkns), or even a collection of scripts in a `Makefile` may be less resource-heavy but allow the same as Dockerfile for checking the code styles :) However the Dockerfile will ensure that the environment is consistent and of course, in that sense, it would also be very beneficial 👍 Maybe a combination of these options will be a nice middle ground, thanks a lot for the suggestion, I really like this proposal.  

I think your suggestion is good, and after rethinking it, I guess pre-commit is the best fit because it's the easiest to set up.  (For an experienced developer like you, these may not be a problem😅, my thoughts are just for your reference)

The reason is that people who use the pypiserver project do not necessarily have experience using docker and makefile because the most basic usage of pypiserver is to use pip install directly. (from QuickStart)

When developers find a requirement, they may be able to contribute to it, and the project's contribute guide can guide them to complete the PR smoothly with the lowest learning cost.
They don't need to understand [makefile syntax](https://stackoverflow.com/questions/28712585/when-to-use-space-or-tab-in-makefile) or docker or even wsl (if their OS is windows).  Sparks won't be doused by frustration with these pre-set-up processes.

PR #445, opened 2022-09-06
fix: force setuptools update + no duplicate runs in GH Actions

dee-me-tree-or-love:
# What

This PR addresses 2 issues with current CI.

1. The failing tests on Python 3.6 due to `setuptools` being out of date (solution roughly based on https://github.com/docker-library/python/issues/170)
2. Currently, for the central repository PRs, GitHub Actions are running double: on `push` and `pull_request` triggers. 

## Changes

1. Forcing `setuptools` update before installing `tox`, seems to fix the problem with tests on Python 3.6.
2. Instead of duplicate runs, like #444, the actions are run only once (see this PR checks).
    - Before:  
      image
    - Now:
      image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,11 @@ name: CI
 on:
   # This will run when any branch or tag is pushed
   push:
-    # standalone is an old branch containing a fully functional pypiserver
-    # executable, from back in the day before docker & a better pip.
-    branches-ignore:
-      - standalone
+    branches:
+      - "master"
     tags:
       - "v**"
-  # Allowing to run on fork pull requests
+  # Allowing to run on fork and other pull requests
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,13 +5,11 @@ name: CI
 on:
   # This will run when any branch or tag is pushed
   push:
-    # standalone is an old branch containing a fully functional pypiserver
-    # executable, from back in the day before docker & a better pip.
-    branches-ignore:
-      - standalone
+    branches:
+      - "master"
```
dee-me-tree-or-love, line 9: @mplanchard just double checking for sanity, this should still respect the ignore that was set on `standalone`, correct? 🙃

dee-me-tree-or-love:
Merging this, but in case of any feedback, please do post comments :v:

PR #466, opened 2022-12-13
feat: get default index url from global pip config

waketzheng:


dee-me-tree-or-love:
Looking at it quickly, I think `pip` by default will fallback to a global PIP index if a package is not found in any of the `--extra-index-url`s (see documentation here: <https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-extra-index-url>). Wouldn't that be sufficient for your use case? 🤔 Let me know!

waketzheng:
@dee-me-tree-or-love pypiserver is running on our ubuntu server for over two years. It works fine all the time, thanks~

PR #470, opened 2023-01-12
Update README.rst and config.py

poshul:
Add note that the units of `--Cache-Control` is seconds.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -187,6 +187,7 @@ not officially supported, and will not receive bugfixes or new features.
       --cache-control AGE
         Add "Cache-Control: max-age=AGE, public" header to package downloads.
         Pip 6+ needs this for caching.
+        AGE is specified in seconds.
 
     pypi-server -h, --help
       Show this help message.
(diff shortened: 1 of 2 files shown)


dee-me-tree-or-love:
Hey @poshul, thanks a lot for this catch! That's a good one, I think this will require a small change in the CLI setup even to make it the most accurate :-) The README docs only aim to copy the output of the CLI's `--help`, so we should fix it there first, and then update the README details. In particular, it should be fixed here: https://github.com/pypiserver/pypiserver/blob/ae3dcf2bbd26f9c282e54327f6ceb0baab1b5420/pypiserver/config.py#L449-L457. Would you be interested in doing this? :-)

poshul:
Thanks. Just updated config.py to match.

dee-me-tree-or-love (approved):
Great work! Thank you so much @poshul! I just briefly resolved a merge conflict and will merge the PR as soon as everything passes :) Sorry for the long wait and thanks again for your addition :)

PR #471, opened 2023-01-30
chore: disable tests on Python3.6

dee-me-tree-or-love:
# What

Since the update of `ubuntu-latest` image in [setup-python](https://github.com/actions/setup-python) GitHub Action to `Ubuntu 22.04`, Python 3.6 is no longer available by default. https://github.com/pypiserver/pypiserver/actions/runs/3861663821/jobs/6683883640#step:3:10. See related issue: https://github.com/actions/setup-python/issues/561.  

Since Python 3.6 is declared "end of life" and is planned to be deprecated for `pypiserver 2.0.0`, I believe it's worth disabling these tests. Related to https://github.com/pypiserver/pypiserver/issues/107 and the discussion with @cclauss in https://github.com/pypiserver/pypiserver/pull/447.

PR #472, opened 2023-01-30
fix: explicit optional types in `config.py`

dee-me-tree-or-love:
# What

This solves [a small typing issue](https://github.com/pypiserver/pypiserver/actions/runs/4048433497/jobs/6963668078#step:5:1) in `config.py`:

```bash
...
Run mypy docker/test_docker.py pypiserver/config.py  tests/test_init.py
pypiserver/config.py:692:48: error: Incompatible default for argument "auther" (default has type "None", argument has type "Callable[[str, str], bool]")  [assignment]
...
pypiserver/config.py:831:48: error: Incompatible default for argument "args" (default has type "None", argument has type "Sequence[str]")  [assignment]
...
```

Solved by explicitly declaring it as `t.Optional`.

### DIFF
--- a/pypiserver/config.py
+++ b/pypiserver/config.py
@@ -828,7 +828,7 @@ class Config:
         return default_config.with_updates(**overrides)
 
     @classmethod
-    def from_args(cls, args: t.Sequence[str] = None) -> Configuration:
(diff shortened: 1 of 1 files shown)


dee-me-tree-or-love:
NB: the failing tests for Python 3.6 will be fixed in https://github.com/pypiserver/pypiserver/pull/471 ✌️

PR #474, opened 2023-02-02
Upgrade to psf/black stable style 2023

cclauss:
@dee-me-tree-or-love Your review, please.

Updating https://github.com/psf/black -> 23.1.0.
* https://github.com/psf/black/blob/main/CHANGES.md#2310

> This is the first [psf/black] release of 2023, and following our [stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy), it comes with a number of improvements to our stable style…

### DIFF
--- a/pypiserver/cache.py
+++ b/pypiserver/cache.py
@@ -14,7 +14,6 @@ try:
     ENABLE_CACHING = True
 
 except ImportError:
-
     Observer = None
 
     ENABLE_CACHING = False
--- a/pypiserver/config.py
+++ b/pypiserver/config.py
@@ -621,7 +621,6 @@ class _ConfigCommon:
         return levels.get(self.verbosity, logging.NOTSET)
 
     def get_backend(self, arg: str) -> IBackend:
-
         available_backends: t.Dict[str, BackendFactory] = {
             "auto": get_file_backend,
             "simple-dir": SimpleFileBackend,
--- a/requirements/dev.pip
+++ b/requirements/dev.pip
@@ -6,6 +6,6 @@
 -r exe.pip
 -r test.pip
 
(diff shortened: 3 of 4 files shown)


dee-me-tree-or-love:
Thanks a lot for following up on this @cclauss! As you noticed I pinned the black version in #471. For me that would make it a little easier to manage the stable integration knowing that the dependencies are specifically defined :) Would you mind bumping the pinned black version in this PR to 23+? Otherwise it's perfect and I'm ready to merge 👍

cclauss:
The discussion in the commit message refers to the `stable style` which means **once a year psf/black will make minor tweeks** to ensure that Python code looks the same on all projects.  Given that this happens only once a year, it is better to stick with the current Python Software Foundation guidelines than to be left behind.

dee-me-tree-or-love:
@cclauss, I fully understand the point and agree with the not-falling-behind idea! :) Given that there will be more changes throughout the year (seeing the releases of `22.X.Y`), and expecting that eventually we'll reach `24.X.Y` in a year from now, I'd personally vote for using a tool like [`dependabot`](https://github.com/dependabot) to maintain the up-to-date versioning in a contained fashion, preventing CI issues on unrelated changes 🤔 So in that case, we'd get notified of the recent updates in a separate PR, allowing us to adapt the code base accordingly when necessary, keeping the other work less impacted. But I am not 100% certain how to set this up yet and I will probably make it a milestone for the next maintenance steps :) What would be your opinion on this? 

Having said this, I do agree that until that point is reached, we can just drop the pin, so let's get this PR merged 👍 :)

dee-me-tree-or-love (approved):
Looking good 👍

Issue #462, opened 2022-11-22
Allow only certain users to upload


Rizhiy:
Hi, I'm trying to run a private pypiserver and I understand that I can restrict access with `-P` and `-a`.
The way I currently understand it is that the same permissions are set for all users.

Is it possible to have different permissions for different users?

I need two types of users:
* "standard" user, which will be able to only list & download/install packages with username & password.
* "admin" user which will be able to upload packages as well.

How can I achieve this?

vEpiphyte:
@Rizhiy I've written a proxy application to do this, that implements a per group and per user access controls for list/download/upload. That sat in front of the simple APIs and filtered out the results as needed. Unfortunately this isn't currently a open source project.

tdg5:
This is functionality I'm looking for too. I'm not against making a pull request, but could use some guidance on how authorizing could be handled per user.

I'm inclined to introduce a new property to the config called `authorizer` and use that authorizer function to handle the per-user authorization. It's a little weird cause it is so similar to the `auther` function, but it seems like a larger breaking change if the `auther` function is updated to take an action in addition to the user name and password it already takes.

More concretely, once an `authorizer` function was available, I'd change [the logic here](https://github.com/pypiserver/pypiserver/blob/e0c9723ee70750d427c34682573efc4b3288da9b/pypiserver/_app.py#L45) to look like this so that both the `auther` and `authorizer` functions would be applied to each request to authenticate/authorize the requested action:
```python
                if (
                    config.auther(*request.auth)
                    and config.authorizer(request.auth[0], request.auth[1], self.action)
                ):
                    return method(*args, **kwargs)
                raise HTTPError(403)
```

I made a draft PR of this notion [here](https://github.com/pypiserver/pypiserver/pull/480), but it is untested and there is still more work to do before the PR would be mergable (e.g. tests)

tdg5:
I made another PR to try to support this feature here: https://github.com/pypiserver/pypiserver/pull/483 . This PR is more of a breaking change, but is simpler then adding another auth function.

ghinks:
Just making an enquiry as this is a feature that I am interested in. Do you want a helping hand with this?

dee-me-tree-or-love:
hey @tdg5 and @ghinks, sorry for quite some silence on this feature proposal! I do really like the idea but didn't get to get to reviewing it yet indeed, sorry about this 😔 If you are both interested, I'd of course really appreciate any help in getting this forward!    

[...]

PR #480, opened 2023-03-10
Add authorizer to authorize which actions a user can take

tdg5:
This relates to https://github.com/pypiserver/pypiserver/issues/462, but is a more general solution that would allow the server to be configured in such a way that any action could be authorized on a per-user basis.

There's not a lot here so far, because I'd like to get general approval on this direction before committing more time to this proposal.

I created a new `authorizer` function because it seemed like changing `auther` to do this work would be a bigger breaking change that maybe wasn't worth it.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -42,9 +42,12 @@ class auth:
                     raise HTTPError(
                         401, headers={"WWW-Authenticate": 'Basic realm="pypi"'}
                     )
-                if not config.auther(*request.auth):
-                    raise HTTPError(403)
-            return method(*args, **kwargs)
+                if (
+                    config.auther(*request.auth)
+                    and config.authorizer(request.auth[0], request.auth[1], self.action)
+                ):
+                    return method(*args, **kwargs)
+                raise HTTPError(403)
 
         return protector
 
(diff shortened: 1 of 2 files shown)


pypiserver/_app.py
```diff
@@ -42,9 +42,12 @@ def protector(*args, **kwargs):
                     raise HTTPError(
                         401, headers={"WWW-Authenticate": 'Basic realm="pypi"'}
                     )
-                if not config.auther(*request.auth):
-                    raise HTTPError(403)
-            return method(*args, **kwargs)
+                if (
+                    config.auther(*request.auth)
+                    and config.authorizer(request.auth[0], request.auth[1], self.action)
```
tdg5, line 47: I've made the `authorizer` also take user name and password because in some implementations the username is a dummy value and the password really has all the info in it. For example, if the password were an access token of some sort, then the username could take on any value and would not be helpful for authorization purposes.

dee-me-tree-or-love:
Hey @tdg5 thanks a lot for your suggestion and PR! This sounds very curious and I am really sorry for the lack of response lately. I will do my best to review this in the near future - I'll add it to my list. For now, I'll need to wrap my head around it, and hope to get back to this shortly! Thanks and sorry for the delays once again.

PR #493, opened 2023-06-20
Feat/dependabot

ghinks:
Add dependabot.yml file. I found that the test.pip file was not recognized and a link to a test-requirements.txt was needed.


dee-me-tree-or-love:
Great work! Thanks a lot for the effort @ghinks and also @cclauss. I like this solution and think we're safe to go with it! If necessary, we can try to find a way around it in a follow-up. This is a very useful addition 👍

dee-me-tree-or-love (approved):

PR #512, opened 2023-08-20
readme output diff on format error

ghinks:
The mdformat markdown checker does not provide a means of displaying what it 
believes the desired formatting should be and the checked in README file.

Therefore:

- Check for the condition when the mdformat finds markdown errors and fails
- On failure create a temp copy of the README update it with the mdformat tool
- Diff the checked in README against the result of the application of the mdformat tool

This PR is a follow up from the conversion from RST to MD of the readme as I did 
not want to add more changes to that already large PR.


dee-me-tree-or-love (approved):
This is awesome, thanks a lot @ghinks 👍

Issue #392, opened 2021-05-16
feature: automatic backoff to avoid too-many-requests for update command


mplanchard:
See #388 for context

dee-me-tree-or-love:
To avoid the rate limiting of XMLRPC, #524 and #410 propose using basic HTTP with JSON parsing. This works very well in my opinion. In the upcoming PR #588 this should be solved (until new HTTP-related issues appear 🙃😁)  :v:

Issue #388, opened 2021-04-21
Command "pypi-server -U" crashes with error HTTPTooManyRequests


nikoluis:
Recently I setup pypi-server on a VM with windows 10 and i have setup all of my Raspberry PIs to have their pip download from this machine. Since i operate on a large scale organization, my only solution to have newest packages, is through a local pypi-server that connects to internet. It works fine, i installed it as a windows service. Today I wanted to download the latest versions of my packages (around 120 packages) so i used pypi-server -U c:\MY_PACKAGES, but i got the following error and any help would be appreciated:

checking 115 packages for newer version

.Traceback (most recent call last):
  File "c:\program files\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\program files\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Program Files\Python39\Scripts\pypi-server.exe\__main__.py", line 7, in <module>
  File "c:\program files\python39\lib\site-packages\pypiserver\__main__.py", line 310, in main
    update_all_packages(
  File "c:\program files\python39\lib\site-packages\pypiserver\manage.py", line 214, in update_all_packages
    update(packages, destdir, dry_run, stable_only)
  File "c:\program files\python39\lib\site-packages\pypiserver\manage.py", line 198, in update
    need_update = find_updates(pkgset, stable_only=stable_only)
  File "c:\program files\python39\lib\site-packages\pypiserver\manage.py", line 120, in find_updates
    pypi_versions = pypi.package_releases(pkg.pkgname)
  File "c:\program files\python39\lib\xmlrpc\client.py", line 1116, in __call__
    return self.__send(self.__name, args)
  File "c:\program files\python39\lib\xmlrpc\client.py", line 1458, in __request
    response = self.__transport.request(
  File "c:\program files\python39\lib\xmlrpc\client.py", line 1160, in request
    return self.single_request(host, handler, request_body, verbose)
  File "c:\program files\python39\lib\xmlrpc\client.py", line 1176, in single_request
    return self.parse_response(resp)
  File "c:\program files\python39\lib\xmlrpc\client.py", line 1348, in parse_response
    return u.close()
  File "c:\program files\python39\lib\xmlrpc\client.py", line 662, in close
    raise Fault(**self._stack[0])
[...]

PR #410, opened 2022-01-07
Replace XMLRPC server usage with JSON API.

royw:
With XMLRPC disabled on pypi.org, pypiserver upgrade no longer works.  This patch replaces the XMLRPC usage for getting package release information with using the pypi.org JSON API and requests library.

Unit tests all pass after this patch.

PR #524, opened 2023-09-01
Use simple JSON API to get package releases.

sjjf:
This replaces the XML-RPC mechanism (which is deprecated and heavily rate-limited, as documented at
https://warehouse.pypa.io/api-reference/xml-rpc.html) with the simple JSON API (PEP 691, https://peps.python.org/pep-0691/).

This is a minimalistic implementation with almost no error handling, but in this context it should behave about the same as the old XML-RPC code (which also had almost no error handling). No additional packages are required, and it should work without issues from Python 3.6 through 3.11 (tested on 3.10)

This should resolve issues #392 and #388.

### DIFF
--- a/pypiserver/manage.py
+++ b/pypiserver/manage.py
@@ -3,23 +3,21 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 import itertools
+import json
 import os
 import sys
-
-from packaging.version import parse as packaging_parse
+from distutils.version import LooseVersion
 from pathlib import Path
 from subprocess import call
-from xmlrpc.client import Server
+from urllib.error import URLError
+from urllib.request import Request, urlopen
 
 import pip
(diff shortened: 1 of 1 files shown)


sjjf:
Just realised this . . . replaces, I guess? . . . #410 - I /think/, if I'm reading the various bits of doco correctly, this would be preferred over the "full" JSON API (see https://warehouse.pypa.io/api-reference/json.html), but this is a maze of twisty passages so I'm not 100% certain about that.

dee-me-tree-or-love:
This is great! Thanks a lot @sjjf! I think this is a good idea, I'll take it up in a separate branch to slightly massage the `diff` and make sure #410 is also referenced and will then merge it to the `master` branch. I hope that's okay with you? Thanks a lot for your contribution! 🌟

PR #459, opened 2022-11-01
FEAT: Switch default hash-algo to sha256

jayeff:
I prepared a PR which switches default hash-algo to sha256 as discussed in #452. Created as draft PR as it probably should not be merged before major version bump to 2.0 is concrete.

Closes #452

Issue #452, opened 2022-10-05
Discussion: Should pypiserver switch default hash algorithm to sha256?


jayeff:
Current default configuration of pypiserver does no longer work with poetry >= 1.2 as pypiserver uses md5 by default for hashing. There is an open issue for this here: https://github.com/python-poetry/poetry/issues/6301 (and workaround with using `--hash-algo=sha256` here: https://github.com/python-poetry/poetry/issues/4523#issuecomment-923741164).

This may still be fixed by a future release of poetry. Still I wonder if it would be useful to switch the default hashing algorithm to sha256 as this is current recommendation by PEP 503

>  Repositories SHOULD choose a hash function from one of the ones guaranteed to be available via the hashlib module in the Python standard library (currently md5, sha1, sha224, sha256, sha384, sha512). **The current recommendation is to use sha256.** 

(note: highlight is mine)

Changing the algorithm would be a breaking change

lucsorel:
🙏 thank you @jayeff for opening this issue.

We have our own instance of pypiserver and would like to upgrade poetry from 1.1.15 to 1.2.0.

If there is a command to update the hashes (from md5 to sha256) in the pypiserver side, I am also interested.

dee-me-tree-or-love:
Hello @jayeff and @lucsorel for bringing this up! 

That's a very important point and I also see the importance of going towards `sha256` as a default algorithm for `pypiserver` too. However, as @jayeff correctly remarks, that would be a breaking change. So I truly hope to include it in the `2.0.0` release along with a [few other new modernisation changes](https://github.com/pypiserver/pypiserver/milestone/11). For now `--hash-algo=sha256` is indeed the best workaround we have at hand, but I would love to move towards `sha256` as default. 

@lucsorel I believe that restarting the `pypiserver` instance with `--hash-algo=sha256` while keeping the same package directories should be out of the box and allow updating the hashes. It will invalidate any caches of course, but then if you generate a Lockfile on the poetry side, it will use the `sha256` hashes provided by the running instance of `pypiserver`. I hope that this is the right approach, and I stand very open to corrections! (I am just starting to dive into this part of `pypiserver`, so maybe I am still missing a few pieces of the puzzle 🙏)  ✌️🙃

lucsorel:
[...]

PR #526, opened 2023-09-08
chore: warn on the default hashing algorithm

dee-me-tree-or-love:
# What

Adding a small warning message when a default hash algorithm is used. 
Just to make sure that the switch to `sha256` (see #459) does not remain unnoticed. 
This change can be reverted later once the change is less likely to be unexpected.

### DIFF
--- a/pypiserver/config.py
+++ b/pypiserver/config.py
@@ -62,6 +62,8 @@ try:
 except ImportError:
     HtpasswdFile = None
 
+log = logging.getLogger(__name__)
+GH_DISCUSSIONS_URL = "https://github.com/pypiserver/pypiserver/discussions"
 
 # The "strtobool" function in distutils does a nice job at parsing strings,
 # but returns an integer. This just wraps it in a boolean call so that we
@@ -115,6 +117,12 @@ def auth_arg(arg: str) -> t.List[str]:
 
 def hash_algo_arg(arg: str) -> t.Optional[str]:
     """Parse a hash algorithm from the string."""
+    if arg is DEFAULTS.HASH_ALGO:
+        log.warning(
+            f"Using the default hashing algorithm: {arg}.\n"
+            + "Please make sure it is consistent with your project setup (e.g. lockfiles).\n"
+            + f"For possible migration strategies, consider {GH_DISCUSSIONS_URL}."
+        )
     if arg in hashlib.algorithms_available:
         return arg
     try:
--- a/pypiserver/config.py
+++ b/pypiserver/config.py
@@ -117,12 +117,6 @@ def auth_arg(arg: str) -> t.List[str]:
 
 def hash_algo_arg(arg: str) -> t.Optional[str]:
     """Parse a hash algorithm from the string."""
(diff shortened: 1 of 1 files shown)


dee-me-tree-or-love:
@jayeff if you have a moment, what do you think? 🤔

jayeff:
@dee-me-tree-or-love I'm not really sure what the purpose for this warning message should be 🤔 

If I interpret the code correct it would only log this warning if I explicitly use the `--hash-algo` argument and it's the same as the default hash algorithm (`sha256` from 2.0 forward). So the essential message would be "your `--hash-algo` argument is redundant because you use the default".

IMO the relevant breaking change is if you use the default hash algorithm pre-2.0 and continue to use the default hash algorithm post-2.0 because then your hashsums from poetry.lock will don't match (see discussion thread on #452) and poetry install no longer works.

I believe your suggested warning does not catch this case

I do question a bit if many people would catch such a warning in the logs anyways (I probably would not 😅). I think clear documentation in the release notes / changelog would be sufficient and probably something more people do check before updating (or check after the update when stuff breaks 😄)

hope this helps

dee-me-tree-or-love:
> @dee-me-tree-or-love I'm not really sure what the purpose for this warning message should be 🤔 
> 
> 
> 
> If I interpret the code correct it would only log this warning if I explicitly use the `--hash-algo` argument and it's the same as the default hash algorithm (`sha256` from 2.0 forward). So the essential message would be "your `--hash-algo` argument is redundant because you use the default".
> 
> 
> 
> IMO the relevant breaking change is if you use the default hash algorithm pre-2.0 and continue to use the default hash algorithm post-2.0 because then your hashsums from poetry.lock don't match (see discussion thread on #452) and poetry install no longer works.
> 
> 
> 
> I believe your suggested warning does not catch this case
> 
> 
> 
> I do question a bit if many people would catch such a warning in the logs anyways (I probably would not 😅). I think clear documentation in the release notes / changelog would be sufficient and probably something more people do check before updating (or check after the update when stuff breaks 😄)
> 
> 
> 
> hope this helps


Hey @jayeff, thanks for the suggestion! That's exactly the second pair of eyes I really was looking for :) 

Ah, I totally agree, thanks for the good catch, I made this as a quick draft without giving this a proper try :D Indeed it's not even going to trigger when necessary 👍 

The motivation behind it was from my recent inspiration by the tools that guide the user through in case of errors - so I wanted to anticipate exactly that possible breaking change item (for people who rely on `latest` tags especially). I still do have a feeling some log statement might help people debug if things unexpectedly broke, but I fully agree that the release notes are even more important 😊

Thanks a lot for thinking along! Then I'll rethink this PR when I'm back to keyboard, but probably I'll focus on the release and troubleshooting notes first indeed.  

Much obliged! 👍

dee-me-tree-or-love:
Closing since it's not a high priority indeed :) Might be a good thing to revisit later 👍

dee-me-tree-or-love:
Closing since it's not a high priority indeed :) Might be a good thing to revisit later 👍

PR #536, opened 2023-10-18
Handle tar.xz archives

doctaweeks:


dee-me-tree-or-love:
hey @doctaweeks, thanks for your suggestion!

That's nice, if I understand correctly, this will add support for `tar.xz` compression on archive sdists, for example [coming from `distutils`](https://github.com/pypa/pip/pull/3204#issuecomment-171675034), right? 😊  


It looks great IMHO and I just have two small requests: 

- Could you please add a small description to this PR so that future folks are able to better discover the reasons for this change? 👾  
- And could you add a test input with `tar.xz` extension here? 😇 https://github.com/pypiserver/pypiserver/blob/2ab30ea2ec7ee52009b1e6daa16e7c5fad3c8722/tests/test_pkg_helpers.py#L8-L16

Thank you! If there won't be any objections, I'll be happy to merge! 😊

dee-me-tree-or-love (approved):
Thanks @doctaweeks, this is looking good and will be a nice addition! :) 
I'll add the test cases myself quickly after. ✌️

PR #546, opened 2024-01-02
Update bottle.py

anonymousdouble:
refactor with fstring to format string to make code more Pythonic.

### DIFF
--- a/pypiserver/bottle.py
+++ b/pypiserver/bottle.py
@@ -3256,8 +3256,8 @@ class BaseTemplate(object):
             if not fname.startswith(spath): continue
             if os.path.isfile(fname): return fname
             for ext in cls.extensions:
-                if os.path.isfile('%s.%s' % (fname, ext)):
-                    return '%s.%s' % (fname, ext)
+                if os.path.isfile(f'{fname}.{ext}'):
+                    return f'{fname}.{ext}'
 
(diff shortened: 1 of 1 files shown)


dee-me-tree-or-love:
Hey @anonymousdouble, thank you for the PR! As of now, we're actually duplicating the `bottle.py` from the original https://github.com/bottlepy/bottle/blob/master/bottle.py of https://github.com/bottlepy/bottle. Maybe this change would be better to direct upstream to `bottle` directly? :) 

I'd prefer to limit possible drifts between our and their sources. 

Additional thought, if you're interested, we could look into making `bottle.py` optionally duplicated. Right now we use it as backend for the `pypi-server` but we don't install it with PIP (historic decision). I do like having a very tiny WSGI server as part of this project directly, but maybe we could look into making it actually an installable dependency. What do you think?

dee-me-tree-or-love:
Hello @anonymousdouble thanks a lot for your PR! With #619 (thanks to @oz123!) we have updated Bottle to the newer version altogether and aim to make it correctly vendored as a regular dependency in the near future.

Thanks a lot for this suggestion and effort again! 
Since we're now running on a newer `bottle`, I'll close this PR, feel free to reopen or reach back if you'd like :v:

PR #547, opened 2024-01-04
Deduplicate host.

OscartGiles:
pypiserver will use `X-Forwarded-Host` to determine the Host for redirects. Because of an odity with nginx ingress this headers is being set twice, and it is not trivially for us to remove the first. Instead we have patched pypiserver to detected when multiple host entries are returned, and only to take the last one.

Fixes https://linear.app/quaisr/issue/Q-1726/cant-install-from-pypi-using-instructions-in-docs


OscartGiles:
Apologies - meant to create this against my fork.

Issue #468, opened 2022-12-31
arm64 flavored image


cmdjulian:
Is it possible to provide an arm64 based image as well? Your Dockerfile seems to be already compatible with it. I checked it out locally and run a multi-arch build in the root of the repository with `docker buildx build -t cmdjulian/pypiserver:v1.5.1 --platform=linux/arm64,linux/amd64 --push .` which worked without any problems.  

I thinks it's just a matter of adding the architecture, in that case arm64, to the github action ci file. It took quite a while, because `building wheels for collected packages: gevent, bcrypt, watchdog, zope.interface` is slow in emulated arm64, but in the end it still worked out just fine and the resulting image is usable without any limitions. You can find it as [cmdjulian/pypiserver:v1.5.1](https://hub.docker.com/r/cmdjulian/pypiserver/tags).  For the future there maybe is a way to speed up the compilation in Docker by using pre-compiled architecture specific packages.

One small hint, for running local multi-arch builds, you have to set up the `qemu binfmt` emulators. For me using arch it was just running `sudo pacman -S qemu-user-static-binfmt`. For ubuntu it's basically the same with `sudo apt install qemu-user-static`. Alternatively you can use `docker run -it --rm --privileged tonistiigi/binfmt --install all` to temporarily add the emulators. This whoever is not persistent throughout reboots.

dee-me-tree-or-love:
Great, thank you for the idea @cmdjulian! I will look into this and hope to add support for ARM architectures with a `v2.0.x` release. I think this is also related to https://github.com/pypiserver/pypiserver/issues/364 and the work of @isac322 in https://github.com/pypiserver/pypiserver/pull/381 seems to address this, so I will see how to approach this!

cmdjulian:
Thanks for looking into it. Do you have any time estimate for the V2 release?

dee-me-tree-or-love:
> Thanks for looking into it. Do you have any time estimate for the V2 release?

Hey @cmdjulian, looking at the current state of things, I'm hoping to get to this within the coming half-year, aiming probably for the summer/early September. I'll try to plan it a bit better/more realistic during the spring months :)

caniko:
Still doesn't exist :(

cbwchuck:
Saw #381 and it has been 3 years 😞...

Open #553 to support amd64 + arm64 only. It took 30 mins for the arm build during my test run on GitHub Actions, with most of the time spent on building wheel for `gevent`.

[...]

PR #553, opened 2024-03-10
Bump github action versions and add multiarch support

cbwchuck:
For addressing #468 

Only added arm64 + amd64 in that I have built and tested them locally and everything seemed to work fine.

Happy to add more archs if needed


dee-me-tree-or-love (approved):
Thanks a lot @cbwchuck for addressing this! Let's give it a try! 🎉

cbwchuck:
Thanks @dee-me-tree-or-love ! 🎉

PR #583, opened 2024-07-04
Update Deprecated Compose Commands for Pypiserver Services

Ykrej:
pypiserver compose services were using the deprecated `pypiserver` command instead of `pypiserver run` command.


dee-me-tree-or-love (approved):
Nice catch, thank you so much @Ykrej!

PR #608, opened 2024-10-01
Update alpine image to 3.20 and python version to 3.10

afdy:
Move to latest alpine image, and later version of python.

This upgrades many packages within the container and fixes many security vulnerabilities.

Python 3.11 fails to build so settled on python 3.10 for now.


dee-me-tree-or-love (approved):
Awesome! Thank you very much @afdy! 🌟

Issue #409, opened 2021-12-31
Supper Cache Package


danerlt:
I set up a local private pypi source, starting with the command `pypi-server -v -p 8080 -a update --log-file /var/log/pypiserver.log -P /opt/pypi/auth/htpasswd.txt --fallback-url https://mirrors.aliyun.com/pypi/simple/ --overwrite /opt/pypi/packages`

If I specify `--fallback-url` as `https://mirrors.aliyun.com/pypi/simple/`, I want to be able to automatically download from the fallback-url to the local `/opt/pypi/packages` directory when I use the pip install command, Can add a --download parameter to implement the caching feature?

PR #621, opened 2024-11-08
Pull through - allow automatic downloads of packages from pypi.org

oz123:
This can have a nice impact for making docker builds locally much faster. Also, for organizations who want faster builds, this is nice.
I imagine that if this is popular enough, this can also have an effect on traffic to pypi.org.

In any case, this still needs to be implemented.  I documented my thougts, how I would do it.
The simple naive solution is implemented in the first commit.
The second commit, explains why the naive solution does not work well, and how I would like to fix it.

Would love to get some feedback here.

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -294,6 +294,20 @@ def simple(project):
     if project != normalized:
         return redirect(f"/simple/{normalized}/", 301)
 
+    # pip install always uses 2 requests, first to get the list of versions
+    # and then to download the package.
+    # If we only search for the package on the local filesystem, we will
+    # might not be able to return the list containing the version.
+    # Hence, if pull_through is enabled, we will fetch the version list
+    # from pypi.org and then return the list of versions.
+
+    # If the packages version is not found, in pypi.org, we will return a 404
+
+    # Serving the actual package is handled by the /packages/ route
+    # And in this route, we will first attempt to fetch the package from the
+    # local filesystem, if not found, we will fetch it from pypi.org and save
+    # it to the local filesystem and then serve it.
+
     packages = sorted(
         config.backend.find_project_packages(project),
         key=lambda x: (x.parsed_version, x.relfn),
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -26,6 +26,10 @@ from .bottle import (
 from .pkg_helpers import guess_pkgname_and_version, normalize_pkgname_for_url
 
 log = logging.getLogger(__name__)
+# your editor may complain about `config` being undefined.
+# To understand why this code work, see how pypiserver/__init__.py
+# imports the _app module and sets the `config` variable in _app.py
+# insidet the function app_from_config.
 config: RunConfig
 app = Bottle()
 
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -313,21 +313,25 @@ def simple(project):
     # it to the local filesystem and then serve it.
     if config.pull_through:
         # Getting a list of packages and hashes
-        rgx = re.compile(r'href=\"(?P<url>.*)#sha256=(?P<meta>.*\") ?>(?P<pkgversion>.*)</a><br />')
-        res = urllib.request.urlopen(f"{config.fallback_url.rstrip('/')}/{project}/")
-        content = res.read().decode('utf-8').split('\n')
+        rgx = re.compile(
+            r"href=\"(?P<url>.*)#sha256=(?P<meta>.*\") ?>(?P<pkgversion>.*)</a><br />"
+        )
+        res = urllib.request.urlopen(
+            f"{config.fallback_url.rstrip('/')}/{project}/"
+        )
+        content = res.read().decode("utf-8").split("\n")
         new_content = []
         for line in content:
             m = rgx.search(line)
             if m:
                 gdict = m.groupdict()
-                vrsn = gdict['pkgversion']
-                meta = gdict['meta']
+                vrsn = gdict["pkgversion"]
+                meta = gdict["meta"]
                 val = f'<a href="http://{config.host}:{config.port}/packages/{vrsn}#{meta}>{vrsn}</a><br />'
             else:
                 val = line
             new_content.append(val)
-        body = '\n'.join(new_content)
+        body = "\n".join(new_content)
         response = HTTPResponse(body=body, status=200)
         return response
 
@@ -406,13 +410,17 @@ def list_packages():
 @auth("download")
 def serve_metadata(pkg_version):
     pkg, version, _ = pkg_version.split("-", 2)
-    urs = urlsplit(config.fallback_url.rstrip('/'))
+    urs = urlsplit(config.fallback_url.rstrip("/"))
     json_url = f"{urs.scheme}://{urs.hostname}/pypi/{pkg}/json"
     response = urllib.request.urlopen(json_url)
-    content = json.loads(response.read().decode('utf-8'))
-    metadata_url = [i['url'] for i in content['releases'][version] if i['packagetype'] == 'bdist_wheel'][0]
-    response = urllib.request.urlopen(metadata_url+".metadata")
-    content = response.read().decode('utf-8')
+    content = json.loads(response.read().decode("utf-8"))
+    metadata_url = [
+        i["url"]
+        for i in content["releases"][version]
+        if i["packagetype"] == "bdist_wheel"
+    ][0]
+    response = urllib.request.urlopen(metadata_url + ".metadata")
+    content = response.read().decode("utf-8")
     response = HTTPResponse(body=content, status=200)
     return response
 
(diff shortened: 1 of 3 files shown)


oz123:
#409 seems like this PR is what the OP wanted.

dee-me-tree-or-love:
Hey @oz123, thanks a lot for your PR! 
I'll try to take as deep of a look as I manage this weekend and leave some comments :) 
Hope to merge it soon (maybe in a week due to some traveling coming up) 👍

pypiserver/_app.py
```diff
@@ -357,25 +406,56 @@ def list_packages():
...
+    content = json.loads(response.read().decode("utf-8"))
+    metadata_url = [
+        i["url"]
+        for i in content["releases"][version]
+        if i["packagetype"] == "bdist_wheel"
+    ][0]
+    response = urllib.request.urlopen(metadata_url + ".metadata")
+    content = response.read().decode("utf-8")
+    response = HTTPResponse(body=content, status=200)
+    return response
```
dee-me-tree-or-love, line 425: [curious] Hey @oz123, do I understand this correctly that this endpoint retrieves `metadata` fully based on the `fallback_url`? Do we not need to consider local packages in this case?

pypiserver/_app.py
```diff
@@ -357,25 +406,56 @@ def list_packages():
...
+        response = static_file(
+            filename,
+            root=str(config.roots[0]),
+            mimetype=mimetypes.guess_type(filename)[0],
+        )
+
+    if config.cache_control:
+        response.set_header(
+            "Cache-Control", f"public, max-age={config.cache_control}"
+        )
```
dee-me-tree-or-love, line 456: [curious] This is really nice. So in this case, we will download the `pull-through` version from a fallback URL and store it as if it was a local package. That's neat 👍  Is that correct?

pypiserver/_app.py
```diff
@@ -289,14 +298,55 @@ def simple(project):
...
+                gdict = m.groupdict()
+                vrsn = gdict["pkgversion"]
+                meta = gdict["meta"]
+                val = f'<a href="http://{config.host}:{config.port}/packages/{vrsn}#{meta}>{vrsn}</a><br />'
+            else:
+                val = line
+            new_content.append(val)
+        body = "\n".join(new_content)
+        response = HTTPResponse(body=body, status=200)
+        return response
```
dee-me-tree-or-love, line 336: [curious] do I correctly understand that in case we've enabled `pull_through`, the simple request returns only the masked URLs from the fallback URL? And then we implicitly rely on retrieving them from the fallback or local. Please correct me if I misunderstand. 

I wonder wouldn't it actually be more valuable to return a union of the URLs: resolved locally U masked from the pull-through setting. What do you think? 

In my opinion we'll then return a more complete set. Iirc, we can't guarantee that versions uploaded locally (suppose someone's pushing forked package versions but falls back to a set remote) and those stored on the fallback are the same? Sort of a `FULL OUTER JOIN` situation :)   

But let me know if I'm mistaken please :v:

pypiserver/_app.py
```diff
@@ -289,14 +298,55 @@ def simple(project):
...
+
     packages = sorted(
         config.backend.find_project_packages(project),
         key=lambda x: (x.parsed_version, x.relfn),
     )
+
     if not packages:
```
dee-me-tree-or-love, line 343: [idea] here I'm also thinking, that to speed up the retrieval of `local U fallback` union of url sets (see comment above), we could run these requests in parallel with something like `asyncio.gather`. But I have to admit I honestly don't recall what is our state of supporting async code inside `pypi-server` right now, I'll need to refresh this 🙈😅

dee-me-tree-or-love (commented):
Hey @oz123 , if I understood the idea well, I think this is a great addition and would indeed be valuable in different use cases. I can imagine using this for sure :)  

I left a few questions to understand the solution best, could you check if I interpret it correctly so we're on the same track? 🙂  

Then we can dive a bit deeper on organising the code and writing some tests for it too. I'll be happy to help out with any parts if you need me :v:

pypiserver/_app.py
oz123: Yes. That is correct. 
Let's say you start with an empty local directory. If you point pip to your local pypiserver and try to install a package foo.
It will not be found, and then it will be fetched with metadata for pypi.org.
If you will try a second time (if there is no local pip cache) the package will be served from pypiserver without first going to get the package from pypi.org.
(Note, you can replace pypi.org with any upstream).

pypiserver/_app.py
oz123: Basically, you are correct. 
If you ask for a package without a version than the correct thing would be to present the client with a union of packages found locally and upstream.
But if you point the client for fetching a package upstream you will not be be able to do the pull-through trick.

(Also, usually, clients do ask for packages with versions)

Example dependancy graph:

A->B->C

You do pip install A.

pypiserver receives a request for A.

A isn't found locally.

Pypiserver fetches the metadata and presenta it pip .

pip determines that version 12.3.4 is the latest version and that there is wheel.
It sends back a request to pypiserver requesting A-12.3.4.whl.

pypiserver does not find it locally so it fetches the file from upstream and saves it locally, then it serves it to pip.

Pip downloads the wheel, installs it, the sees that it needs to ask for B in version 9.8.7. The cycle goes on again ...,

pypiserver/_app.py
oz123: As per my response above the process is pretty much sequential and fetching the packages using an async client isn't going to be faster.

We can do better by shortening the cycle by not having to asked pypi.org for metadata every time by caching it. But for that I'd like to do a new PR introducing a better storage backend...

pypiserver/_app.py
dee-me-tree-or-love: Right, not that sounds all good. What I mean is more regarding this: 

> If you ask for a package without a version than the correct thing would be to present the client with a union of packages found locally and upstream.

> But if you point the client for fetching a package upstream you will not be be able to do the pull-through trick.

The part I am concerned about is not the masking of the urls as I understand that in the pull-through case, we'll eventually fetch them too. What I mean is that here it seems we return (line 336) _only_ the remote URLs masked, without the local packages at all. 

So suppose package `A` has: 
1. locally: `A:0.0.1`, `A:0.0.2`, and ``A:0.0.2-rc1`
2. remotely: `A:0.0.1`, `A:0.0.2`, and `A:0.0.3`

Then in my opinion the output of the `/simple/A` should be as follows: 
- `A:0.0.1` 
- `A:0.0.2`
- `A:0.0.2-rc1`
    > this is only on local
- `A:0.0.3` 
    > this appears as if on local, but will be pulled-through from the remote

Afaiu, current code with `pull-through` enabled, will return only the `A:0.0.1`, `A:0.0.2`, and `A:0.0.3`. Is that correct?
Does it make sense? 🤔 

Let me know :v:

oz123:
@syvasha that is an idea, I would like to implement this as a storage backend which is actually a "NullStorage".

My plan is to make this feature such that it is built in a way that one can specify pull-through:storagebackend pairs.
That way, you can also create an "ExpiringBackend" to trim older versions of a package.

Let me know what you think.

s1eke:
May I ask about the progress of this Pull Request?

syvasha:
Hi, I'm sorry for a late reply, got caught up with some things.

> I would like to implement this as a storage backend which is actually a "NullStorage".
> 
> My plan is to make this feature such that it is built in a way that one can specify pull-through:storagebackend pairs. That way, you can also create an "ExpiringBackend" to trim older versions of a package.

Do I understand correctly that you would like to have it possible to have two different storage backends for the two package groups? (Those directly uploaded to pypiserver vs. those that were pulled through from PyPI) That sounds like a good idea!

As for the ExpiringBackend, it would be a neat feature - but maybe it is not necessary? Except for handling pulled-through packages. Otherwise, I would imagine it would be unlikely that an organization would want to prune their "own" packages after some time (unless they produce so much it is comparable to the resource load of having all the pulled through packages stored); one can always build again from source, but some would, for the sake of simplicity, prefer to have their "archived" versions in one place, without setting up e.g. a second pypiserver instance. But I guess its a matter of preference and allocated time haha

Pruning older packages, or, say, regular pruning of 'dev' builds, can be done by a cron job on the filesystem that stores the package files (e.g. we have a weekly script that cleans up all .devX packages); I guess you can leave that up to the end users.

The only difficulty I see there is differentiating between intentionally uploaded packages and those that were pulled through - if they can be separated in the storage in such a way that external scripts/services can easily differentiate, then it would be easy to solve with a cron job.

If you have time to implement the expiring backend, though, and make it able to differentiate between pulled-through and intentionally uploaded packages, I'm sure it will be a popular feature :)
the NullStorage backend for pulled-through packages would also be a solution, a simpler one, and also great!

Cheers

adouge:
Hey :) I guess you did not have time to work on this more. Is there anything one could help you with? Would really like to see this feature

upd. In the meantime, in case anyone is looking for a solution to forwarding PyPI to closed local networks, setting up a limited local PyPI mirror with `pypa/bandersnatch`, that can then be used as a secondary index alongside `pypiserver`, seems like a feasible option

PR #539, opened 2023-11-13
fix: support Python 3.12

dee-me-tree-or-love:
# What

This PR makes sure `pypi-server` is compatible with Python 3.12. 

## Changes

- [x] Pin the latest supported version to Python 3.12
- [x] For certainty, test the 'stable' Python `3.x` in `test-python` 
- [x] Solve the dependency issues
  - [x] Backward-compatible [migration](https://peps.python.org/pep-0632/#migration-advice) from deprecated `distutils`
  - [x] Backward-compatible `pkg_resources`
  - [x] Fix the docker test issues
- [x] Make sure the CI is passign

## Getting involved? :)

All help is very welcome 😁 If you are interested, feel free to chime in in this PR and/or open forks with partial steps pointing to this `fix-add-python-3-12-support` branch. All formats are very much appreciated ✌️

PR #627, opened 2024-11-24
fix: use `importlib_resources` only when needed

dee-me-tree-or-love:
# What

We removed the upper bound for `importlib_resources`.
However, this package is required for the temporary support in Python version 3.8-3.12 only (see #539 when it was originally introduced).

## Changes

- return `importlib_resources` upper bound in `setup.py`
- fix the `envlist` of `tox.ini` to be aligned with `ci.yaml`


mgorny:
Thank you!

PR #649, opened 2025-06-03
Remove obsolete known limitation relating to misspellings

pelson:
As highlighted in https://github.com/pypiserver/pypiserver/issues/38#issuecomment-226874194, PEP-503 specified a normalisation scheme to enable redirects in this case. PyPI/Warehouse does not redirect common misspellings outside of normalisation.


dee-me-tree-or-love:
Hey @pelson, thanks a lot for noticing this and submitting the PR! Good catch, thank you for the fix!

PR #675, opened 2025-11-19
Remove support for Python 3.8 & 3.9

harryzcy:
Python 3.8 and 3.9 are EOL now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ options:
   --welcome HTML_FILE   Use the contents of HTML_FILE as a custom welcome
                         message on the home page.
   --cache-control AGE   Add "Cache-Control: max-age=AGE" header to package
-                        downloads. Pip 6+ requires this for caching.AGE is
+                        downloads. Pip 6+ requires this for caching. AGE is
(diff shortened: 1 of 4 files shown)


dee-me-tree-or-love (approved):
Great catch! Thanks for following up 👍

README.md
```diff
@@ -47,14 +47,14 @@ making it much easier to get a running index server.
 
 Table of Contents
 
-- [**pypiserver - minimal PyPI server for use with pip/easy_install**](#pypiserver---minimal-pypi-server-for-use-with-pipeasy_install)
+- [**pypiserver - minimal PyPI server for use with pip/easy\_install**](#pypiserver---minimal-pypi-server-for-use-with-pipeasy_install)
```
dee-me-tree-or-love: As we use the standard `mdformat` configuration, it requires `easy_install` without a `_`. Do you mind adjusting?

We should probably fix this in the formatter's config some day!

harryzcy: @dee-me-tree-or-love I see that you pushed up a commit to fix those. Thanks!

PR #691, opened 2026-04-03
feat: add upload-time metadata for uv exclude-newer

kimdwkimdw:
This change makes pypiserver expose the upload-time metadata that `uv --exclude-newer` needs in order to make correct time-based candidate decisions for packages hosted by pypiserver.

Closes #690

## Summary

- persist upload-time metadata when files are uploaded
- attach upload-time to discovered packages, with mtime fallback when metadata is absent
- expose `files[].upload-time` and `versions` from `/simple/<project>/` JSON responses
- add an end-to-end `twine -> pypiserver -> uv --exclude-newer` test
- bump the Docker base image to Python 3.12 / Alpine 3.23

## Why

Public PyPI already exposes file-level upload timestamps, but private packages hosted by pypiserver did not. That meant `uv exclude-newer` could not apply the same cutoff policy to internal artifacts served by pypiserver.

This patch closes that gap with the smallest Simple API surface needed for uv.

## Test plan

- `pytest tests/test_upload_time.py tests/test_backend.py tests/test_app.py -v`
- `pytest tests/test_server.py -k twine_upload_json_upload_times_drive_uv_exclude_newer -vv`
- `python -m tox -e py`

### DIFF
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -88,7 +88,8 @@ def simple_project_versions(packages):
 
 
 class auth:
-    """decorator to apply authentication if specified for the decorated method & action"""
+    """decorator to apply authentication if specified for the decorated
+    method & action"""
 
     def __init__(self, action):
         self.action = action
@@ -97,7 +98,10 @@ class auth:
         def protector(*args, **kwargs):
             if self.action in config.authenticate:
                 if not request.auth or request.auth[1] is None:
-                    raise HTTPError(401, headers={"WWW-Authenticate": 'Basic realm="pypi"'})
+                    raise HTTPError(
+                        401,
+                        headers={"WWW-Authenticate": 'Basic realm="pypi"'},
+                    )
                 if not config.auther(*request.auth):
                     raise HTTPError(403)
             return method(*args, **kwargs)
@@ -439,7 +443,10 @@ def server_static(filename):
                 mimetype=mimetypes.guess_type(filename)[0],
             )
             if config.cache_control:
-                response.set_header("Cache-Control", f"public, max-age={config.cache_control}")
+                response.set_header(
+                    "Cache-Control",
+                    f"public, max-age={config.cache_control}",
+                )
             return response
 
     return HTTPError(404, f"Not Found ({filename} does not exist)\n\n")
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -79,6 +79,7 @@ def simple_project_file_json(current_uri, package):
         "filename": os.path.basename(package.relfn),
         "url": simple_project_url(current_uri, package),
         "hashes": simple_project_hashes(package),
+        "size": os.path.getsize(package.fn),
         "upload-time": format_upload_time(package.upload_time),
     }
 
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -26,10 +26,17 @@ from .bottle_wrapper import (
     template,
 )
 from .pkg_helpers import guess_pkgname_and_version, normalize_pkgname_for_url
+from .upload_time import format_upload_time
 
 log = logging.getLogger(__name__)
 config: RunConfig
 app = Bottle()
+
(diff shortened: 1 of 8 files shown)


dee-me-tree-or-love:
Hey @kimdwkimdw, this is a great idea! Thanks a lot for the PR. 

I'm a little under time pressure, so to help review this, would you mind adding some explanatory comments over your diffs? That'd really help navigating the changes a bunch 😇 🙏

kimdwkimdw:
Hi @dee-me-tree-or-love 

Sorry about the extra noise in the earlier version of this PR. I accidentally let some lint-only formatting changes slip in, which made the review harder than it needed to be.

I’ve rolled those back in `ee2744c`, so the remaining diffs should stay focused on the actual behavior change.

To make the current diff easier to navigate, the changes are now basically grouped around a few main pieces:
- persist upload-time metadata in a sidecar file when a package is uploaded, and clean it up again when a package is deleted
- attach `upload_time` to discovered package entries, with mtime as a fallback when metadata is missing
- expose the minimum Simple API JSON fields that `uv --exclude-newer` needs from `/simple/<project>/`
- verify the behavior end-to-end with `twine -> pypiserver -> uv --exclude-newer`, while trimming back some of the test helper noise

I also added `8f0e49a`, which fixes the upload-time metadata lifecycle and reduces some of the review overhead in the new server test.

Thanks again for taking a look.

yangfan-yf-yf (changes requested):
The JSON response declares Simple API version 1.4, but each file object currently omits the mandatory `size` field. On `4213ad1`, the response contains `filename`, `url`, `hashes`, and `upload-time`; the Simple Repository API requires `size` to be an integer for every file from API version 1.1 onward.

Because `versions` and `upload-time` also require at least version 1.1, lowering the declared version would not make this response conformant. Could you add the package file size to each file object and cover it in the JSON response regression?

I ran the focused upload-time, backend, and app suites on this head (`181 passed`). They remain green, which confirms the current tests do not catch the missing required field.

kimdwkimdw:
Thanks @yangfan-yf-yf, you’re right.

PEP 700 makes `files[].size` mandatory for JSON Simple API version 1.1 and later, so the 1.4 response was incomplete.

I’ve added the artifact size in bytes using `os.path.getsize(package.fn)`.

yangfan-yf-yf (changes requested):
Thanks for adding the required `files[].size` field. I verified the response change on this head and reran the upload-time, backend, and app tests (`181 passed`).

The new end-to-end test still has two Windows portability blockers:

1. `run_twine()` passes Windows paths through `shlex.split()`. This turns a path such as `C:\Users\...\.pypirc` into `C:Users...pypirc`; my run therefore stopped in Twine with `InvalidConfiguration` before uploading anything. Please build the argv list directly and pass `str(conf)` and `str(package)` as separate elements.
2. The test later hard-codes `venv_dir / "bin" / "python"`, while `uv venv` creates `Scripts/python.exe` on Windows. Please select the environment interpreter for the current platform.

Since pypiserver advertises Windows support, could we keep this regression test portable before merging?

kimdwkimdw:
@yangfan-yf-yf 

Fixed both Windows compatibility issues by passing Twine arguments as an argv list and using Scripts/python.exe for Windows virtual environments. The focused upload/register tests pass locally.

yangfan-yf-yf (approved):
I rechecked 505fe0bb2d1cdf74bed52f8efc27c36c1791de7f on Windows. Both portability issues from my previous review are resolved.

With Python 3.12.14 and UTF-8 enabled for subprocess output (`PYTHONUTF8=1`, `PYTHONIOENCODING=utf-8`), `pytest tests/test_server.py -k twine -q` passes all seven selected tests. This includes real local Twine uploads and the uv `--exclude-newer` regression: the persisted upload timestamps survive restart and reversed mtimes, JSON reports file sizes, and uv selects the expected older/newer versions. I used an isolated home and temporary directory for the local test server and `.pypirc`.

Without UTF-8 output, my GBK console encountered a separate Twine/Rich UnicodeEncodeError; the successful rerun above used UTF-8 explicitly.

PR #704, opened 2026-05-29
fix: harden /RPC2 XML parser against entity-expansion DoS (#685)

gistrec:
## Summary

Fixes #685 (CWE-776, billion-laughs / entity-expansion DoS in the unauthenticated `/RPC2` search endpoint).

The endpoint parsed request bodies with `xml.dom.minidom.parse()`, which the [Python stdlib docs explicitly mark as unsafe](https://docs.python.org/3/library/xml.html#xml-vulnerabilities) against maliciously constructed input. Combined with the default-anonymous `list` action (the `@auth(\"list\")` decorator only enforces auth when `\"list\"` is in `config.authenticate`, and the default is `[\"update\"]`), a ~700-byte DTD payload triggers either an Expat amplification error (HTTP 500) on Expat ≥ 2.4.0 or uncapped exponential entity expansion (OOM) on older Expat — and many long-lived base images still ship Expat 2.2.x.

## Fix

- Switch from `xml.dom.minidom` to `defusedxml.minidom`, which rejects internal entity bombs, external entities (XXE), and DTD-based attacks by default.
- Convert the resulting `DefusedXmlException` / `ExpatError` into a clean HTTP 400 instead of an unhandled 500.
- Add `defusedxml >= 0.7.1` as a runtime dep. `defusedxml` is the canonical XML-hardening library recommended by the Python stdlib docs, ~25 KB, pure-Python, stable since 2013, no further transitive deps.

I also considered a no-new-dep variant (rejecting `<!DOCTYPE` / `<!ENTITY` at the wire — valid since XML-RPC doesn't use DTDs). Happy to switch to that on request if the dep is undesirable; defusedxml seemed the cleaner default since it's the standard fix recommended for exactly this CWE.

## Reproduction

Before this change (Python 3.11 / Expat 2.7.3):

```bash
curl -X POST http://localhost:8080/RPC2 \
  -H 'Content-Type: text/xml' \
  --data @level7-billion-laughs.xml
# HTTP/1.1 500 Internal Server Error
# ExpatError: limit on input amplification factor (from DTD and entities) breached
```

On a Python linked against Expat < 2.4.0 (e.g. RHEL 8 with system Expat 2.2.5, or many older container images), the same request triggers full exponential expansion → OOM.

After this change: HTTP 400 with body `Invalid or unsafe XML payload`, returned in microseconds without any expansion attempt.

## Tests

- Added `test_rpc_rejects_unsafe_or_invalid_xml` covering three payloads:
  - billion-laughs DTD (the original PoC, shortened),
  - XXE attempting `file:///etc/passwd` exfiltration,
  - plain malformed XML.
- Existing `test_search` (parametrized, 6 cases) still passes — no regression on the legitimate search code path.

```
tests/test_app.py ........................................................... 169 passed
```

(The `tests/test_server.py` integration suite has pre-existing failures unrelated to this change — `python -m build --wheel --no-isolation` returns 256 on `main` as well.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 3.0.0 (tbd)
 -----------
 
+- FIX: security: harden ``/RPC2`` XML parser against entity-expansion DoS
+  ("billion laughs", CWE-776). Switch from ``xml.dom.minidom`` to
+  ``defusedxml.minidom`` and reject malformed/unsafe XML payloads with
+  HTTP 400 (#685).
+
 2.4.1 (2026-02-10)
 --------------------------
 
--- a/pypiserver/_app.py
+++ b/pypiserver/_app.py
@@ -2,13 +2,16 @@ import logging
 import mimetypes
 import os
 import re
-import xml.dom.minidom
 import xmlrpc.client as xmlrpclib
 import zipfile
 from collections import defaultdict, namedtuple
 from io import BytesIO
 from json import dumps
 from urllib.parse import quote, urljoin, urlparse
+from xml.parsers.expat import ExpatError
+
+import defusedxml.minidom
+from defusedxml.common import DefusedXmlException
 
 from pypiserver.config import RunConfig
 
@@ -233,7 +236,10 @@ def pep_503_redirects(project=None):
 @auth("list")
 def handle_rpc():
     """Handle pip-style RPC2 search requests"""
-    parser = xml.dom.minidom.parse(request.body)
+    try:
+        parser = defusedxml.minidom.parse(request.body)
+    except (DefusedXmlException, ExpatError):
+        raise HTTPError(400, "Invalid or unsafe XML payload")
     methodname = (
         parser.getElementsByTagName("methodName")[0]
         .childNodes[0]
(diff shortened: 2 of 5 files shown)


dee-me-tree-or-love:
Thanks a lot for the fix @gistrec! Very neat change and I really appreciate the added tests. 

I agree `defusedxml` is a good move and in line with community recommendations. And thanks for highlighting an alternative approach. 

I'll give it all a closer look coming workdays and merge it unless there'll be something to discuss 👍

Thanks again!

dee-me-tree-or-love:
This looks great @gistrec 👍  Merging ✅ 

An idea for the future: I was looking up more about XML vulnerabilities and found this PR https://github.com/python/cpython/pull/135294. From which it seems that XML parsing in Python 3.11+ is safe by default. So later on, it might be interesting to experiment with running the tests with default XML on Python 3.11+. In the style of this discussion in another repo: https://github.com/jdepoix/youtube-transcript-api/pull/596. But I don't think it's urgent, just thinking out loud. 

Thanks for the fix again!

dee-me-tree-or-love (approved):
Nice 👍

Issue #703, opened 2026-05-25
Bug: not working when hosted under a reverse proxy with URL prefix


h3har:
**Version:** `pypiserver==2.4.1`

There are multiple issues in `pypiserver` that prevent it from functioning properly when hosted under a reverse proxy with a URL prefix:
 - Redirects do not consistently use the proper server base URL (via `server_base_url` or the `X-Forwarded-*` headers)
     - Redirects for normalized package names in `_app.py` use relative URLs, which breaks when hosted under a prefix.
 - The common HTTP reverse proxy headers don't appear to be handled properly
     - `X-Forwarded-Prefix` (`HTTP_X_FORWARDED_PREFIX` via WSGI) is ignored, which requires an "abuse" of `X-Forwarded-Host` (`HTTP_X_FORWARDED_HOST` via WSGI) to contain the full URL (host + prefix), rather than just the host name.
     - This is less of an issue and would require a breaking change to fix, but it is something that appears suboptimal.

## Code excerpts

Normalized redirects using relative URLs:
```python
# PEP 503: require normalized project
normalized = normalize_pkgname_for_url(project)
if project != normalized:
    return redirect(f"/simple/{normalized}/", 301)
```
```python
# PEP 503: require normalized project
normalized = normalize_pkgname_for_url(project)
if project != normalized:
    return redirect(f"/{normalized}/json", 301)
```

PR #719, opened 2026-08-01
FIX: preserve URL prefixes in normalized redirects

neallin917:
## Summary

- build normalized project redirects from the current prefix-aware request path
- preserve mounted/server base URLs for both Simple API and JSON redirects
- add regression tests for both endpoints under a mounted `/priv` prefix
- document the fix in the 3.0.0 changelog

## Root cause

The normalized-name handlers returned root-absolute paths such as `/simple/foo-bar/` and `/foo-bar/json`. When the application was mounted below a URL prefix, those redirects escaped the mount point.

The updated handlers reuse `request_fullpath(request)` and replace only the project portion, preserving the current mount path. This addresses the concrete normalized-redirect problem in #703 without changing forwarded-header semantics.

## Validation

- `uv run --locked --group test pytest tests/test_app.py -k 'normalized_name_redirect' -vv` — 12 passed
- `uv run --locked --group test pytest tests` — 497 passed
- `uv run --locked --group dev ruff format --check pypiserver/_app.py tests/test_app.py`
- `git diff --check`

## AI assistance disclosure

OpenAI Codex was used to investigate the issue, draft the implementation and regression tests, and run the validation commands listed above.


yangfan-yf-yf (approved):
Verified current head `26709382` against #703. Both affected handlers now derive the redirect base from `request_fullpath(request)` and replace only the project segment. This preserves `SCRIPT_NAME`/`--server-base-url` mounts and the repository's existing path-in-`X-Forwarded-Host` handling, while leaving root-mounted redirects unchanged. The new regression tests cover both the Simple API and JSON endpoints under `/priv`, and the existing parametrized cases continue to cover all supported normalization forms.

The normalized-redirect selection passes (12 tests), `tests/test_app.py` passes (171 tests), and `git diff --check` is clean. The eight failures in the broader Windows run are confined to `tests/test_server.py` pip/twine subprocess and server cases and do not exercise these handlers. I found no overlapping open implementation. This looks correct.

dee-me-tree-or-love:
@neallin917 thanks for the PR! We went for #725 over this. I'll mention your effort in the release in any case 👌

Thanks for the AI use disclosure 🙏

PR #725, opened 2026-08-10
fix: preserve proxy prefixes in normalization redirects

### LINKED ISSUE #703: Bug: not working when hosted under a reverse proxy with URL prefix

Neallin-917:
## Summary
- build PEP 503 and JSON normalized-name redirects from the current request path instead of the host root
- preserve URL prefixes supplied by `X-Forwarded-Host` and by the server-base-url mount
- add regression coverage for both supported deployment forms

Fixes #703

## Validation
- `uv run pytest tests/test_app.py -k normalized_name_redirect -q` (14 passed)
- `uv run pytest -q`
- `uv run ruff format --check pypiserver/_app.py tests/test_app.py`
- `git diff --check`

The repository-wide Ruff and mypy invocations report existing diagnostics in these legacy dynamic Bottle modules (for example global `config` injection and missing annotations); this PR does not change those areas.

## Disclosure
AI assistance was used while preparing this contribution; I reviewed the change and ran the validation above.


yangfan-yf-yf (commented):
I compared this head with #719. Both change the same two normalized-name redirect branches for #703, so they cannot both merge independently.

The current-path `urljoin` approach here preserves both supported deployment forms, and the expanded regressions cover `X-Forwarded-Host` path prefixes as well as `--server-base-url`. The normalized-redirect selection passes (`14 passed`), and the complete `tests/test_app.py` file passes (`173 passed`). #719 already carries the changelog entry and has a completed review.

Could you consolidate the two pull requests into one selected branch, retaining these broader proxy-prefix tests and the changelog entry, then close the other one? I will re-review the consolidated head afterward.

Neallin-917:
Thanks. I consolidated the selected implementation into this PR:

- retained the broader `X-Forwarded-Host` and `--server-base-url` prefix coverage
- added the requested `CHANGES.rst` entry for #703
- reran the normalized-redirect selection: 14 passed

The overlapping PR was opened from a previous account that I no longer operate, so I cannot close it directly. Please close #719 if you agree that #725 is the selected branch. I am ready for re-review here.

yangfan-yf-yf (approved):
I rechecked `a054bd7` against current `main`. The consolidated branch now retains the broader proxy-prefix and server-base-url regressions and includes the #703 changelog entry.

The normalized-redirect selection passes (`14 passed`), and the complete `tests/test_app.py` file passes (`173 passed`). The worktree is clean after the run, and I found no technical blocker on this head.

Because #719 remains open and its author cannot close it, a maintainer still needs to select this branch and close #719 before merge. From the code and test side, LGTM.

Neallin-917:
Friendly follow-up: #725 retains the proxy-prefix coverage and changelog entry, and now has a technical approval. Since the overlapping #719 cannot be closed by its previous contributor, could a maintainer select this branch and close the duplicate when convenient? Thank you.

dee-me-tree-or-love (approved):
Nice catch! Thank you 👍

Thanks for the engaged review @yangfan-yf-yf

PR #732, opened 2026-09-19
Fix /:project/json download URLs escaping the base URL prefix

### LINKED ISSUE #593: package url in json format is incorrect
this line 
https://github.com/pypiserver/pypiserver/blob/d68355a3712192324b65dc0a5fbe092113e8fb16/pypiserver/_app.py#L396

```python
should be rewritten as {"url": urljoin(req_url, "../packages/" + x.relfn)}
```

There is one level of ../ too much as we come from route "/:project/json"

niukanen1:
The `/:project/json` endpoint built each release's download URL with `urljoin(req_url, "../../packages/" + relfn)`. The route is only one level deep, so `../../` walks one directory too far: at the root the extra `..` is silently absorbed, but when the app is mounted under a prefix (`--server-base-url=/pypi`, or behind a reverse proxy that forwards the prefix) the resulting URL escapes the prefix and points outside the server.

Use `../packages/`, which matches the route depth and keeps the URL under the mount point in both cases.

Fixes #593


dee-me-tree-or-love (approved):
thanks, this is a great catch! 👍

